PR #11, opened 2016-06-13
add functions

double-y:
# Adding text mining functions

Probably the code is not clear yet, so please feel free to put comments how it can be more clear.
I'm worried if the tests are enough, so please be critical if edge cases are considered.

Test

* devtools::test()


R/util.R
```diff
@@ -11,3 +10,108 @@ col_name <- function(x, default = stop("Please supply column name", call. = FALS
...
 }
+
+#' Simple cast wrapper
+simple_cast <- function(data, row, col, val, fun.aggregate=mean, fill=0){
+  loadNamespace("reshape2")
+  fml <- as.formula(paste(row, col, sep = "~"))
+  data %>%  reshape2::acast(fml, value.var=val, fun.aggregate=fun.aggregate, fill=fill)
+}
+
+#' Gather only upper half
```
hideaki: Some more explanation on what "upper half" means would help.

R/pairwise.R
```diff
@@ -0,0 +1,89 @@
...
+  loadNamespace("tidytext")
+  group_col <- col_name(substitute(group))
+  dimention_col <- col_name(substitute(dimention))
+  value_col <- col_name(substitute(value))
+
+  grouped_column <- grouped_by(df)
+
+  cnames <- avoid_conflict(grouped_column, c("pair.name.1", "pair.name.2", "value.sim"))
+
+  # this is executed to each group
```
hideaki: on each group, or over each group would probably sound more natural.

R/util.R
```diff
@@ -11,3 +10,108 @@ col_name <- function(x, default = stop("Please supply column name", call. = FALS
     return(x)
   stop("Invalid column specification", call. = FALSE)
 }
+
+#' Simple cast wrapper
```
hideaki: Some more comments on what this does would help.

R/pairwise.R
double-y: changed it to on

R/util.R
double-y: add explanation

R/util.R
double-y: add explanation

R/util.R
```diff
@@ -11,3 +10,108 @@ col_name <- function(x, default = stop("Please supply column name", call. = FALS
...
+    df <- data.frame(
+      Var1=r_names[ind[,1]],
+      Var2=c_names[ind[,2]],
+      value=val)
+    colnames(df) <- cnames
+    df
+  }
+}
+
+#' group by other columns
```
hideaki: some more explanation on what this function does would help.

R/util.R
```diff
@@ -11,3 +10,108 @@ col_name <- function(x, default = stop("Please supply column name", call. = FALS
...
+group_exclude <- function(df, ...){
+  loadNamespace("dplyr")
+  cols <- as.character(substitute(list(...)))[-1]
+  excluded <- setdiff(colnames(df), cols)
+  # exclude list column
+  target <- excluded[!sapply(df[,excluded], is.list)]
+  dplyr::group_by_(df, .dots=target)
+}
+
+#' prevent name conflict
```
hideaki: some more explanation on what this function does would help.

R/util.R
double-y: add explanation

R/util.R
double-y: add explanation

R/util.R
hideaki: That's good explanation to have too, but I meant the comment as a whole on the subroutine that starts from line 73.

double-y: add explanation there too

PR #17, opened 2016-06-20
upgrade to 0.2.5

double-y:
Change names of functions

Test
* devtools:test()
* use functions installing this to the libpath in .exploratory dir


R/model_builder.R
```diff
@@ -3,18 +3,29 @@
 
 #' Create do wrapper function with source data
 #' @return do wrapper function
-do_data <- function(funcname) {
-  ret <- function(df, ..., keep.source = FALSE){
+build_data <- function(funcname) {
+  ret <- function(df, ..., keep.source = FALSE, augment=F){
```
hideaki: Isn't FALSE and F same thing? Should we uniformly use only one of those?

R/model_builder.R
```diff
@@ -23,78 +34,144 @@ do_data <- function(funcname) {
...
+      %>%  dplyr::do_(.dots=setNames(list(~build_kmeans_each(.), ~(.)), c(model_column, source_column)))
+    )
+    # Add a class for Exploratyry to recognize the type of .source.data
+    class(output[[source_column]]) <- c("list", ".source.data")
+  } else {
+    output <- (
+      df
+      %>%  dplyr::do_(.dots=setNames(list(~build_kmeans_each(.)), model_column))
+    )
+    if(augment){
```
hideaki: Why is this if(augment) case only in under else and not under if(keep.source)?
I guess keep.source implies augment at the next step? Anyway some comment would help.

double-y: Yes, augment should be out from there.
I'll change the first condition to `keep.source | !augment` and put out the augment condition from the else block

R/model_builder.R
double-y: Yes, I'll change them to FALSE

PR #20, opened 2016-06-25
upgrade to 0.2.7

double-y:
* change do_dist.cols to calculate distances between columns
* add do_cmdscale
* change get_sentiment to word_to_sentiment
* add get_sentiment that uses sentimentr

Test
* devtools::test()


R/tools.R
```diff
@@ -30,3 +30,100 @@ generate_examples <- function(function_name, args){
   print(knitr::kable(ret, format="markdown"))
   ret
 }
+
+#' Check package update from cran and requiredRpackage json
```
hideaki: I think this function is about our internal process and should not be shared to public.

R/stats_wrapper.R
```diff
@@ -175,3 +175,49 @@ do_svd.kv <- function(df,
...
+  loadNamespace("dplyr")
+  loadNamespace("tidyr")
+  pair1_col <- col_name(substitute(pair_name1))
+  pair2_col <- col_name(substitute(pair_name2))
+  value_col <- col_name(substitute(value))
+  grouped_col <- grouped_by(df)
+
+  name_col <- avoid_conflict(grouped_col, "name")
+
+  do_cmdscale_each <- function(df){
```
hideaki: I guess this is for group_by again, but some comment would help.

double-y: add a comment

R/tools.R
double-y: This will be moved to our private repository

PR #21, opened 2016-06-28
[WIP] modify result from "nrc" so that it returns list

double-y:
words in nrc lexicon are not unique, so it causes error when used in mutate.

Change it to return list

Test
* devtools::test()


double-y:
this is merged to yy_upgrade0.2.8

PR #33, opened 2016-07-12
mat_to_df sort issue fix

double-y:
If mat_to_df took a matrix with unsorted names, the result was wrong.

* fix mat_to_df not to sort the names inside
* fix do_cor.cols and do_dist.cols so that it sort the name beforehand
* add a test to check if simple_cast makes a matrix with sorted names for .kv functions

And there was a missing , in DESCRIPTION, so fixed it too

Test
* devtools::test()

### DIFF
--- a/R/pairwise.R
+++ b/R/pairwise.R
@@ -122,6 +122,8 @@ do_dist.cols <- function(df, ..., label=NULL, fill=0, fun.aggregate=mean, distin
   # this is executed on each group
   calc_dist_each <- function(df){
     mat <- df %>%  dplyr::select_(.dots=select_dots) %>%  as.matrix()
+
+    # sort the column name so that the output of pair.name will be sorted
     sortedNames <- sort(colnames(mat))
     mat <- t(mat)
     mat <- mat[sortedNames, ]
--- a/R/pairwise.R
+++ b/R/pairwise.R
@@ -123,7 +123,8 @@ do_dist.cols <- function(df, ..., label=NULL, fill=0, fun.aggregate=mean, distin
   calc_dist_each <- function(df){
     mat <- df %>%  dplyr::select_(.dots=select_dots) %>%  as.matrix()
 
-    # sort the column name so that the output of pair.name will be sorted
+    # sort the column name so that the output of pair.name.1 and pair.name.2 will be sorted
+    # it's better to be sorted so that heatmap in exploratory can be triangle if distinct is TRUE
     sortedNames <- sort(colnames(mat))
     mat <- t(mat)
     mat <- mat[sortedNames, ]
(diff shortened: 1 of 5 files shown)


R/pairwise.R
```diff
@@ -122,9 +122,12 @@ do_dist.cols <- function(df, ..., label=NULL, fill=0, fun.aggregate=mean, distin
   # this is executed on each group
   calc_dist_each <- function(df){
     mat <- df %>%  dplyr::select_(.dots=select_dots) %>%  as.matrix()
+    sortedNames <- sort(colnames(mat))
```
hideaki: Could you add comment on why we are sorting here?

R/stats_wrapper.R
```diff
@@ -67,6 +67,8 @@ do_cor.cols <- function(df, ..., use="pairwise.complete.obs", method="pearson",
   # check if the df's grouped
   do_cor_each <- function(df){
     mat <- dplyr::select_(df, .dots = select_dots) %>%  as.matrix()
+    mat <- mat[,sort(colnames(mat))]
```
hideaki: Please add comment on why we are sorting here.

R/pairwise.R
double-y: added a comment

R/stats_wrapper.R
double-y: added a comment

R/pairwise.R
```diff
@@ -123,8 +123,13 @@ do_dist.cols <- function(df, ..., label=NULL, fill=0, fun.aggregate=mean, distin
   calc_dist_each <- function(df){
     mat <- df %>%  dplyr::select_(.dots=select_dots) %>%  as.matrix()
 
+    # sort the column name so that the output of pair.name will be sorted
```
hideaki: I think this comment is a little too terse. It took some time for me to understand what it meant.
- You could say "pair.name.1 and pair.name.2" instead of saying "pair.name".
- Does it make sense to explain why it is important or useful that pair.name.1 and pair.name.2 is sorted? If so, please do.

R/stats_wrapper.R
```diff
@@ -67,6 +67,9 @@ do_cor.cols <- function(df, ..., use="pairwise.complete.obs", method="pearson",
   # check if the df's grouped
   do_cor_each <- function(df){
     mat <- dplyr::select_(df, .dots = select_dots) %>%  as.matrix()
+    # sort the column name so that the output of pair.name will be sorted
```
hideaki: Please see my other comment on the same kind of comment.

R/pairwise.R
double-y: Add comments about it

R/stats_wrapper.R
double-y: Add comments about it

PR #34, opened 2016-07-12
add is_empty

double-y:
add is_empty function

Test
* devtools::test()


R/value_check.R
```diff
@@ -0,0 +1,4 @@
+#' @export
```
hideaki: Could you add comment on what this function does?

PR #37, opened 2016-07-14
modify list_extract so that it can take minus value

yosuke-yasuda:
Modify list_extract to deal with minus position value

Test
- devtools::test()

### DIFF
--- a/R/util.R
+++ b/R/util.R
@@ -184,30 +184,41 @@ list_extract <- function(column, position = 1, rownum = 1){
   }
 
   if(is.data.frame(column[[1]])){
-    sapply(column, function(column){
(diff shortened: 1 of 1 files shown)


R/util.R
```diff
@@ -178,10 +178,36 @@ list_n <- function(column){
...
 list_extract <- function(column, position = 1, rownum = 1){
+
+  if(position==0){
+    stop("position 0 is not supported")
+  }
+
   if(is.data.frame(column[[1]])){
-   sapply(column, function(column) column[rownum, position])
+    sapply(column, function(column){
+      if(position<0 & position >= -ncol(column)){
```
hideaki: Can you put if statement outside of sapply so that it does not have to get repeated?

PR #51, opened 2016-08-01
modify order of arguments of do_tfidf

double-y:
Fix https://github.com/exploratory-io/tam/issues/2681

Test:
* devtools::test()


double-y:
This is aborted because in next steps, the main thing tends to be documents, so it should be regarded as subject

PR #54, opened 2016-08-07
modify .kv functions so that they show better error

double-y:
.kv functions emits "No data found" if a grouped column is chosen for subject_column, so this adds a validations for that cases

Test
devtools::test()

### DIFF
--- a/R/model_builder.R
+++ b/R/model_builder.R
@@ -85,6 +85,10 @@ build_kmeans.kv <- function(df,
   model_column <- avoid_conflict(grouped_column, "model")
   source_column <- avoid_conflict(grouped_column, "source.data")
 
+  if(row_col %in% grouped_column){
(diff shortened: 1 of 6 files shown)


R/model_builder.R
```diff
@@ -85,6 +85,10 @@ build_kmeans.kv <- function(df,
   model_column <- avoid_conflict(grouped_column, "model")
   source_column <- avoid_conflict(grouped_column, "source.data")
 
+  if(row_col %in% grouped_column){
+    stop(paste0(row_col, " is a gruoping column. You can use ungroup() to solve this."))
```
hideaki: Typo: "gruoping".
"You can use..." does not sound very well as error message. Maybe something like following?
XXX is a grouping column. ungroup() may be necessary before this operation.

double-y: modified the error message

PR #55, opened 2016-08-08
add fun.aggregate to cosine similarity

double-y:
Add fun.aggregate to cosine similarity (for sparse matrix)

Test
devtools::test()

### DIFF
--- a/R/pairwise.R
+++ b/R/pairwise.R
@@ -1,12 +1,13 @@
 #' Calculate similarity of each pair of groups.
 #' @param df data frame in tidy format
-#' @param group A column you want to calculate the correlations for.
-#' @param dimension A column you want to use as a dimension to calculate the correlations.
+#' @param subject A column you want to calculate the correlations for.
+#' @param key A column you want to use as a dimension to calculate the correlations.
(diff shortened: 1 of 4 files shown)


R/util.R
```diff
@@ -29,6 +29,40 @@ simple_cast <- function(data, row, col, val, fun.aggregate=mean, fill=0){
   data %>%  reshape2::acast(fml, value.var=val, fun.aggregate=fun.aggregate, fill=fill)
 }
 
+sparse_cast <- function(data, row, col, val=NULL, fun.aggregate=sum){
```
hideaki: Please add comment on what this function does as a whole.

R/pairwise.R
```diff
@@ -8,7 +8,7 @@
 #' @param diag If similarity between itself should be returned or not.
 #' @param method Type of calculation. https://cran.r-project.org/web/packages/proxy/vignettes/overview.pdf
```
hideaki: Looks like the comments here are out of sync with current arguments of do_cosine_sim.kv.
I don't see group, dimension, and method arguments.

PR #57, opened 2016-08-23
add functions for statistical test

double-y:
Add
* do_t.test
* do_var.test
* remove build_t.test
* remove build_var.test

Test
* devtools::test()


R/test_wrapper.R
```diff
@@ -0,0 +1,86 @@
...
+  }
+  # otherwise, one sample t-test from value_col
+
+  grouped_col <- grouped_by(df)
+
+  model_col <- avoid_conflict(grouped_col, "model")
+
+  # this is executed on each group
+  do_t.test_each <- function(df, ...){
+    if(with_key){
```
hideaki: some explanation on what is done when with_key is true or false would help.

double-y: added comments

PR #61, opened 2016-09-04
functions update

double-y:
* make value column argument optional in kv functions
* add score function to replace augment


R/util.R
```diff
@@ -25,20 +25,32 @@ simple_cast <- function(data, row, col, val, fun.aggregate=mean, fill=0){
...
+    mat[mat == 0] <- fill
+    mat
+  }else{
+    data %>%  reshape2::acast(fml, value.var=val, fun.aggregate=fun.aggregate, fill=fill)
+  }
 }
 
 #' Cast data to sparse matrix by choosing row and column from a data frame
-sparse_cast <- function(data, row, col, val=NULL, fun.aggregate=sum){
+sparse_cast <- function(data, row, col, val=NULL, fun.aggregate=sum, count = FALSE){
```
hideaki: Please add some comment on what the count argument does.

R/broom_wrapper.R
```diff
@@ -71,3 +71,18 @@ augment_kmeans <- function(df, model, data){
     }
   })
 }
+
+#' augment wrapper
+#' @export
+score <- function(df, model, ...){
```
hideaki: We will be calling this one predict.

R/util.R
double-y: I put a comment

R/broom_wrapper.R
double-y: Changed to predict

PR #62, opened 2016-09-06
update model functions

double-y:
* fix https://github.com/exploratory-io/tam/issues/2677
* default keep.source TRUE


double-y:
Test
* devtools::test()

PR #64, opened 2016-09-11
urltools wrapper functions

double-y:
Add urltools wrapper functions

Test:
* devtools::test()


NAMESPACE
```diff
@@ -53,6 +53,9 @@ export(getTwitter)
 export(getTwitterToken)
 export(get_sentiment)
 export(get_stopwords)
+export(get_url_param)
+export(get_url_subdomain)
+export(get_url_suffix)
```
hideaki: URL suffix could mean different things like this one.
https://survey.qut.edu.au/WebHelp/URLSuffix.htm

Top-Level Domain would be most unambiguous word here.
I know it's a little long but how about get_url_top_level_domain ?

kanaugust: I think suffix is good enough. See this doc.
https://cran.r-project.org/web/packages/urltools/urltools.pdf

kanaugust: And there is another concept tld (top level domain), which is different from suffix. Having said that, we need to add get_tld for the tld_extract function from urltools. Also, to be consistent we should also add our wrapper functions for the followings.

- domain - get_url_domain
- fragment - get_url_fragment
- parameters - get_url_parameters
- path - get_url_path
- port - get_url_port
- scheme - get_url_scheme

- get_url_param
- get_url_subdomain
- get_url_suffix
- get_url_tld

double-y: Added those functions

R/url_operation.R
```diff
@@ -1,20 +1,77 @@
...
 }
 
 #' get subdomain from url string
 #' @export
 get_url_subdomain <- function(url) {
+  loadNamespace("urltools")
   urltools::suffix_extract(urltools::url_parse(url)$domain)$subdomain
 }
 
+#' get suffix from url string
```
hideaki: The comment should be "get top-level domain from url string"

double-y: Fixed it

PR #69, opened 2016-09-16
modify cluster column in build_kmeans

double-y:
Cluster column should be always integer (if augment=TRUE and .cols, it was factor)
Change column names ".cluster" to "cluster" when augment=TRUE and .kv

Test:
devtools::test()

### DIFF
--- a/R/broom_wrapper.R
+++ b/R/broom_wrapper.R
@@ -41,7 +41,9 @@ augment_kmeans <- function(df, model, data){
   ret <- tryCatch({
     # use do.call to evaluate data_col from a variable
     augment_func <- get("augment", asNamespace("broom"))
-    do.call(augment_func, list(df, model_col, data=data_col))
+    ret <- do.call(augment_func, list(df, model_col, data=data_col))
+    ret[[ncol(ret)]] <- as.integer(ret[[ncol(ret)]])
+    ret
   },
   error = function(e){
     loadNamespace("dplyr")
@@ -86,6 +88,6 @@ predict <- function(df, model, ...){
   if(any(class(df[[model_col]]) %in% ".model.kmeans")){
     augment_kmeans(df, model, ...)
   } else {
-    augment(df, model, ...)
+    broom::augment(df, model, ...)
   }
 }
--- a/R/broom_wrapper.R
+++ b/R/broom_wrapper.R
@@ -42,6 +42,7 @@ augment_kmeans <- function(df, model, data){
     # use do.call to evaluate data_col from a variable
     augment_func <- get("augment", asNamespace("broom"))
     ret <- do.call(augment_func, list(df, model_col, data=data_col))
+    # cluster column is factor labeled "1", "2"..., so convert it to integer to avoid confusion
     ret[[ncol(ret)]] <- as.integer(ret[[ncol(ret)]])
     ret
   },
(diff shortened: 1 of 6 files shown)


R/broom_wrapper.R
```diff
@@ -41,7 +41,9 @@ augment_kmeans <- function(df, model, data){
   ret <- tryCatch({
     # use do.call to evaluate data_col from a variable
     augment_func <- get("augment", asNamespace("broom"))
-    do.call(augment_func, list(df, model_col, data=data_col))
+    ret <- do.call(augment_func, list(df, model_col, data=data_col))
+    ret[[ncol(ret)]] <- as.integer(ret[[ncol(ret)]])
```
hideaki: Please leave comment about why we are casting this to integer.

R/model_builder.R
```diff
@@ -176,7 +176,10 @@ build_kmeans.cols <- function(df, ...,
     mat <- dplyr::select_(df, .dots=select_dots) %>% as.matrix()
     kmeans_ret <- kmeans(mat, centers = centers, iter.max = 10, nstart = nstart, algorithm = algorithm, trace = trace)
     if(augment){
-      broom::augment(kmeans_ret, df)
+      ret <- broom::augment(kmeans_ret, df)
+      colnames(ret)[[ncol(ret)]] <- avoid_conflict(grouped_column, "cluster")
+      ret[[ncol(ret)]] <- as.integer(ret[[ncol(ret)]])
```
hideaki: Please leave comment about why we are casting this to integer.

R/broom_wrapper.R
double-y: added a comment

R/model_builder.R
double-y: added a comment

R/model_builder.R
```diff
@@ -102,15 +102,39 @@ build_kmeans.kv_ <- function(df,
...
   }
 
   build_kmeans_each <- function(df){
     mat <- simple_cast(df, row_col, col_col, value_col, fun.aggregate = fun.aggregate, fill=fill)
-    kmeans_ret <- kmeans(mat, centers = centers, iter.max = 10, nstart = nstart, algorithm = algorithm, trace = trace)
+    kmeans_ret <- tryCatch({
+      kmeans(mat, centers = centers, iter.max = 10, nstart = nstart, algorithm = algorithm, trace = trace)},
+      error = function(e){
+        if(e$message == "cannot take a sample larger than the population when 'replace = FALSE'"){
+          # this matrix falls into here when centers = 3 and mat is
```
hideaki: I appreciate having example, but this example confuses me.
Is it an error because there are only group1 and group2 while centers is 3,
or is it because there are only 1 and 5 in the matrix cells while centers is 3,
or for some other reason?

R/model_builder.R
```diff
@@ -159,41 +179,69 @@ build_kmeans.cols <- function(df, ...,
   loadNamespace("tidyr")
   loadNamespace("broom")
   set.seed(seed)
-
   select_dots <- lazyeval::lazy_dots(...)
   grouped_column <- grouped_by(df)
   model_column <- avoid_conflict(grouped_column, "model")
   source_column <- avoid_conflict(grouped_column, "source.data")
-  selected_column <- setdiff(colnames(dplyr::select_(df, .dots=select_dots)), grouped_column)
+  selected_column <- evaluate_select(df, .dots=select_dots, grouped_column)
```
hideaki: What is this line above doing? A comment would help.

R/model_builder.R
double-y: I'm not completely sure in which condition this error appears but as my experiments, the column numbers doesn't seem to have effect, so I modified the comment accordingly

R/model_builder.R
double-y: Added a comment

PR #70, opened 2016-09-19
update column names

double-y:
Change output column names for kv functions


double-y:
Test
- [x] devtools::test()

PR #72, opened 2016-09-20
svd na handling

double-y:
add an error for NA value in svd

Test:
- [x] devtools::test()

### DIFF
--- a/R/stats_wrapper.R
+++ b/R/stats_wrapper.R
@@ -169,6 +169,9 @@ do_svd.kv <- function(df,
   # this is executed on each group
   do_svd_each <- function(df){
     matrix <-simple_cast(df, subject_col, dimension_col, value_col, fun.aggregate = fun.aggregate, fill=fill)
+    if(any(is.na(matrix))){
+      stop("NA is not supported as value.")
+    }
     if(centering){
       # move the origin to center of data
       matrix <- sweep(matrix, 2, colMeans(matrix), "-")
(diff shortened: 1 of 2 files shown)


R/stats_wrapper.R
```diff
@@ -169,6 +169,9 @@ do_svd.kv <- function(df,
   # this is executed on each group
   do_svd_each <- function(df){
     matrix <-simple_cast(df, subject_col, dimension_col, value_col, fun.aggregate = fun.aggregate, fill=fill)
+    if(any(is.na(matrix))){
+      stop("NA is not supported as value.")
```
hideaki: Looks like this is true for all of Subject, Key, and Value.
Would something like following make it easier to understand?
"NA is not supported as value for Subject, Key, or Value."

tests/testthat/test_stats_wrapper.R
```diff
@@ -91,6 +91,18 @@ test_that("test do_cor without val", {
   expect_equal(result[[3]][1:2], c(1, -1))
 })
 
+test_that("do_svd.kv with NA value", {
+  data <- data.frame(row = c(1, 1, 2, 2, 3,3),
+                     col = rep(c(1,2), 3),
+                     val = seq(6)) %>%
+    dplyr::slice(-3)
```
hideaki: I didn't know this trick of slice with negative number. I would appreciate a comment.

double-y: added comment

PR #73, opened 2016-09-20
modify test wrapper error

double-y:
Test
- [x] devtools::test()

### DIFF
--- a/R/test_wrapper.R
+++ b/R/test_wrapper.R
@@ -25,7 +25,7 @@ do_t.test <- function(df, value, key=NULL, ...){
         t.test(data=df, fml, ...)
       }, error = function(e){
         if(e$message == "grouping factor must have exactly 2 levels"){
-          stop("grouping column has to have 2 unique values")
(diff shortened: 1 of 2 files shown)


R/test_wrapper.R
```diff
@@ -21,7 +21,14 @@ do_t.test <- function(df, value, key=NULL, ...){
   do_t.test_each <- function(df, ...){
     if(with_key){
       # use formula (`value_col`~`key_col`) for two sample t-test
-      model <- t.test(data=df, fml, ...)
+      model <- tryCatch({
+        t.test(data=df, fml, ...)
+      }, error = function(e){
+        if(e$message == "grouping factor must have exactly 2 levels"){
+          stop("grouping column has to have 2 unique values")
```
hideaki: Would it make sense to make grouping column title case, like Grouping Column so that it matches with what we show on command UI?

double-y: I see. The label of the form is Group Column, so I'll use it

PR #81, opened 2016-10-05
modify list_concat so that it can take more than 2 lists

double-y:
### Description

modify list_concat so that it can take columns

ref: https://github.com/exploratory-io/tam/issues/3157

### Checklist

- [x] Pass devtools::check()
- [x] Pass devtools::test() 
- [x] Tested with Exploratory

### DIFF
--- a/R/util.R
+++ b/R/util.R
@@ -347,7 +347,6 @@ list_to_text <- function(column, sep = ", "){
 #' concatinate vectors in a list
 #' @export
 list_concat <- function(..., collapse = TRUE){
-  browser()
   lists <- list(...)
 
   # size of each list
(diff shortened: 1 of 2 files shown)


R/util.R
```diff
@@ -346,8 +346,27 @@ list_to_text <- function(column, sep = ", "){
 
 #' concatinate vectors in a list
 #' @export
-list_concat <- function(list){
-  list(unlist(list))
+list_concat <- function(..., collapse = TRUE){
```
hideaki: With the analogy to paste(), I think it makes more sense to make default FALSE.
I would guess breaking backward compatibility on this function now should have only minimum impact yet.

double-y: Okay, modified to FALSE

PR #87, opened 2016-10-16
HP vertica support

double-y:
### Description
Add HP Vertica support for getDBConnection

### Checklist

- [x] Pass devtools::check()
- [x] Pass devtools::test() 
- [x] Tested with Exploratory


R/system.R
```diff
@@ -274,11 +274,11 @@ getDBConnection <- function(type, host, port, databaseName, username, password){
   if(!requireNamespace("DBI")){stop("package DBI must be installed.")}
   drv = NULL
   conn = NULL
-  if(type == "mysql" || type == "aurora"){
+  if(type == "mysql" | type == "aurora"){
     drv <- DBI::dbDriver("MySQL")
     conn = RMySQL::dbConnect(drv, dbname = databaseName, username = username,
                              password = password, host = host, port = port)
-  } else if (type == "postgres" || type == "redshift"){
+  } else if (type == "postgres" | type == "redshift" | type == "hp_vertica"){
```
hideaki: Let's call the type "vertica" for consistency with others. (It's "redshift" and not "amazon_redshift".)

hideaki: Also, are you sure this "||" should be replaced with "|"?
I thought "|" was for vector operation.

double-y: Okay, I was always using "|" because it also work on one element vector but using "||" seems more apropriate if we are sure that it's one length vector. I'll use || here.
http://stackoverflow.com/questions/6558921/r-boolean-operators-and

I'll change the name too

PR #90, opened 2016-10-17
add a function to get a number of rows in the target mongodb collection

kei51e:
### Description

Added a function `getMongoCollectionNumberOfRows` to get a number of rows in the target mongodb collection. 

### Checklist

- [x] Pass devtools::check()
- [x] Pass devtools::test() 
- [x] Tested with Exploratory


R/system.R
```diff
@@ -254,6 +249,19 @@ queryMongoDB <- function(host, port, database, collection, username, password, q
       url = stringr::str_c(url, "?authSource=", authSource)
     }
   }
+  return (url)
+}
+
+
+#' @export
+queryMongoDB <- function(host, port, database, collection, username, password, query = "{}", isFlatten, limit=100000, isSSL=FALSE, authSource=NULL, fields="{}", sort="{}", skip=0){
```
hidekoji: could you set `limit=0` as a default ? LIn2 wants to use (all i.e. 0) as a default value.

kei51e: I changed it to 0, and tested on exploratory verified it works.

PR #92, opened 2016-10-20
add remote data import functions

double-y:
### Description
Add functions for 
* google trends data import
* twitter timeline data import

### Checklist

- [x] Pass devtools::check()
- [x] Pass devtools::test() 
- [x] Tested with Exploratory


R/twitter_timeline.R
```diff
@@ -0,0 +1,20 @@
+#' Get a user's timeline
+#' @export
+getTwitterTimeline <- function(user, n=3200, includeRts = FALSE,
```
hideaki: What is the reason for 3200 for default n? Please leave the reason as a comment.

R/google_trends.R
```diff
@@ -0,0 +1,83 @@
+#' access google trends
+#' @export
+getGoogleTrends <- function(user, password, query = "", type = "trend", last = "5y", geo = ""){
```
hideaki: Comments on parameters would help especially the possible values of last.

R/twitter_timeline.R
double-y: added comments for arguments

R/google_trends.R
```diff
@@ -0,0 +1,89 @@
+#' Access google trends
+#' @param user - User id.
+#' @param password - Password for the account.
+#' @param query - Vector of queries.
+#' @param type - Output type. Can be "top_regions", "top_cities" or "trends".
+#' @param last - From when the data should be retreived.
```
hideaki: Could you add list of possible values ( "1h", "5y" etc.) so that user of this function does not need to go through the code for the list?

double-y: added them

PR #95, opened 2016-10-22
[Google BigQuery] make sure to use bucket project id instead of query target project id

hidekoji:
### Description
Make sure to use bucket project id instead of target project id.

### Checklist

- [ ] Pass devtools::check()
- [ ] Pass devtools::test() 
- [ ] Tested with Exploratory

### DIFF
--- a/R/system.R
+++ b/R/system.R
@@ -281,7 +281,7 @@ queryMongoDB <- function(host, port, database, collection, username, password, q
   }
 }
 
-#' Returns the total number of rows stored in the target table. 
+#' Returns the total number of rows stored in the target table.
 #' At this moment only mongdb is supported.
 #' @export
 getMongoCollectionNumberOfRows <- function(host, port, database, username, password, collection, isSSL=FALSE, authSource=NULL){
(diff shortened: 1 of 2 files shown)


R/system.R
```diff
@@ -628,24 +628,24 @@ executeGoogleBigQuery <- function(project, sqlquery, destination_table, page_siz
   # if bucket is set, use Google Cloud Storage for extract and download
   if(!is.null(bucket) && !is.na(bucket) && bucket != ""){
     # destination_table looks like 'exploratory-bigquery-project:exploratory_dataset.exploratory_bq_preview_table'
-    dataSetTable = stringr::str_split(stringr::str_replace(destination_table, stringr::str_c(project,":"),""),"\\.")
+    dataSetTable = stringr::str_split(stringr::str_replace(destination_table, stringr::str_c(bucketProjectId,":"),""),"\\.")
```
hideaki: It seems that bucketProjectId info is passed in both bucketProjectId argument and as part of destination_table. Maybe it is for backward compatibility?
Does it make sense to extract bucketProjectId from destination_table rather than adding bucketProjectId argument?

hidekoji: That's for backward compatibility, to make this API with TAM plugin frame, we need to add new argument.

hideaki: OK, so for backward compatibility, I guess this function needs to work even without new arguments like bucketProjectId?
If so, do we need to take care of the case where bucketProjectId is NULL?

PR #103, opened 2016-10-27
update google trends output

double-y:
### Description
Keyword columns should be always from lowered and space removed query

### Checklist

- [x] Pass devtools::check()
- [x] Pass devtools::test() 
- [x] Tested with Exploratory

### DIFF
--- a/R/google_trends.R
+++ b/R/google_trends.R
@@ -76,17 +76,31 @@ getGoogleTrends <- function(user, password, query = "", type = "trend", last = "
     # So the first result is enough
     key <- keys[startsWith(keys, "Top.regions") | startsWith(keys, "Top.subregions")][[1]]
     # gather columns except for the first column (names of region) to make it easy to visualise
+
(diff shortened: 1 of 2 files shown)


R/google_trends.R
```diff
@@ -76,17 +76,31 @@ getGoogleTrends <- function(user, password, query = "", type = "trend", last = "
     # So the first result is enough
     key <- keys[startsWith(keys, "Top.regions") | startsWith(keys, "Top.subregions")][[1]]
     # gather columns except for the first column (names of region) to make it easy to visualise
+
+    # if query is single, geo is used for a column name, so convert it to the query
```
hideaki: The comment is too terse to convey the idea of what is going on.
You could write something like,

If query consists of only one word, gtrendsR uses regions (e.g. Japan) for output column name instead of the query word.
So we are replacing it with the query word so that we can use the query word in subsequent steps.

double-y: Updated that way

PR #106, opened 2016-11-12
improve word_to_sentiment

double-y:
### Description
`word_to_sentiment` was very slow for grouped data frames because it loads mapping data from tidy text for sentiment in each execution (if it's grouped, the function is executed for each groups).

Also, using named vector is faster than join for mapping, so it is used now

This solves it by saving the data to this pacakge internally using `R/sysdata.rd`

The data is created by this command

```
data("sentiments", package = "tidytext", envir = environment())

# only this uses summarize because nrc has many sentiment types, so list is used for one to many relationships among words and sentiments
# the other two is one to one relationships, so named vector can be used
sentiment_nrc_df <- sentiments %>% filter(lexicon == "nrc") %>% group_by(word) %>% dplyr::summarize(sentiment=list(sentiment))
sentiment_nrc <- sentiment_nrc_df[["sentiment"]]
names(sentiment_nrc) <- sentiment_nrc_df[["word"]]

sentiment_bing_df <- sentiments %>% filter(lexicon == "bing")
sentiment_bing <- sentiment_bing_df[["sentiment"]]
names(sentiment_bing) <- sentiment_bing_df[["word"]]

sentiment_AFINN_df <- sentiments %>% filter(lexicon == "AFINN")
sentiment_AFINN <- sentiment_AFINN_df[["score"]]
names(sentiment_AFINN) <- sentiment_AFINN_df[["word"]]

devtools::use_data(sentiment_nrc, sentiment_bing, sentiment_AFINN, internal = TRUE)
```

### Checklist

- [x] Pass devtools::check()
- [x] Pass devtools::test() 
- [x] Tested with Exploratory


R/string_operation.R
```diff
@@ -48,20 +48,20 @@ word_to_sentiment <- function(words, lexicon="bing"){
...
-      )
-  } else {
-    sentiments <- sentiments[check,]
-  }
-  joined_df <- dplyr::left_join(data.frame(word=words, stringsAsFactors = FALSE), sentiments, by="word")
-  if(lexicon=="AFINN"){
-    joined_df$score
-  } else {
-    joined_df$sentiment
+    sentiment_df <- sentiments[check,] %>%
```
hideaki: Please add comment on why this group_by/summarize is needed only for nrc case, and what this is doing.

R/string_operation.R
```diff
@@ -48,20 +48,20 @@ word_to_sentiment <- function(words, lexicon="bing"){
...
+      dplyr::group_by(word) %>%
+      dplyr::summarize(sentiment=list(sentiment))
+    sentiment <- sentiment_df[["sentiment"]]
+
+  } else if (lexicon == "AFINN") {
+    sentiment <- sentiments[check,][["score"]]
+  }else {
+    sentiment <- sentiments[check,][["sentiment"]]
   }
+  names(sentiment) <- sentiments[check,][["word"]]
```
hideaki: Looks like this part is instead of join. Please add comment on what we are doing here, and why this is the better way than join.

tests/testthat/test_string_operation.R
```diff
@@ -35,6 +35,19 @@ test_that("test word_to_sentiment", {
...
+test_that("test word_to_sentiment to groupd_df", {
+  df <- data.frame(
+    text = c("good", "sad", letters[1:(10000 * 3 - 2)]),
+    group = rep(seq(10000), 3),
+    stringsAsFactors = FALSE
+  )
+
+  ret <- df %>%
+    dplyr::group_by(group) %>%
+    dplyr::mutate(sent = word_to_sentiment(text))
```
hideaki: Should we test summarize after group_by too?

R/string_operation.R
double-y: Actually, I realised that this operation can be done when creating internal data, so removed it

R/string_operation.R
double-y: Mentioned join in a comment

PR #108, opened 2016-11-14
[Remote Excel support] Wrapper API for readxl::read_excel and excel_sheets

hidekoji:
### Description

Created wrapper API for

a) readxl::read_excel
b) readxl::excel_sheets

Update following API for remote RDATA files support

c) exploratory::getObjectListFromRdata
d) exploratory::getObjectFromRdata

### Checklist

- [ ] Pass devtools::check()
- [ ] Pass devtools::test() 
- [ ] Tested with Exploratory


R/system.R
```diff
@@ -1089,17 +1089,41 @@ select_columns <- function(x, ...) {
   return (df)
 }
 
+clear_cach_file <- function(url){
```
hideaki: Typo "cach".
Also, some comment about how this is used (to cache downloaded remote file etc.),
how it works (uses tempfile https://stat.ethz.ch/R-manual/R-devel/library/base/html/tempfile.html , and a R variable with name of hashed url is assigned to the path given by tempfile.)

R/system.R
```diff
@@ -1089,17 +1089,41 @@ select_columns <- function(x, ...) {
...
+  hash <- digest::digest(url, "md5", serialize = FALSE)
+  tryCatch({
+    do.call(rm, c(as.name(hash)),envir = .GlobalEnv)
```
hideaki: I think it is better to call unlink() to remove the file. The tempfile page says that is the way to remove file.

R/system.R
```diff
@@ -1089,17 +1089,41 @@ select_columns <- function(x, ...) {
...
 download_excel_file <- function(url){
-  ext <- stringr::str_to_lower(tools::file_ext(url))
-  # if no extension, assume the file extension as xlsx
-  if(ext == ""){
-    ext = "xlsx"
+  shouldCacheExcel <- getOption("tam.should.cache.excel")
+  filepath <- NULL
+  hash <- digest::digest(url, "md5", serialize = FALSE)
+  tryCatch({
+   filepath <- eval(as.name(hash))
```
hideaki: Indent is off here.

R/system.R
```diff
@@ -1089,17 +1089,41 @@ select_columns <- function(x, ...) {
...
+  }, error = function(e){
+    filepath <- NULL
```
hideaki: Some comment would help on this error case about what kind of possible error this is handling.

R/system.R
```diff
@@ -1089,17 +1089,41 @@ select_columns <- function(x, ...) {
...
+  })
+  if(!is.null(shouldCacheExcel) && isTRUE(shouldCacheExcel) && !is.null(filepath)){
+    filepath
```
hideaki: I guess this is cache hit case, but some comment would help to be able to figure it out quickly.

R/system.R
hidekoji: fixed indent.

R/system.R
hidekoji: added comment.

PR #112, opened 2016-11-19
modify gtrends for package update

double-y:
### Description
The output is changed by gtrendR upgrade, so modified the function accordingly

When type is "trend", keywords were in column names but now gtrends put them in keyword column

The amount of access to the keyword is in "hits"column by gtrendsR  but "trend" was used as the column name in exploratory package, so "trend" is still used here

Even though query is more than one, the column names need to be replaced when type is "trend" because Japanese character becomes strange if it's in the column

### Checklist

- [ ] Add test cases for this fix/enhancement - No test because this needs account information
- [x] Pass devtools::check()
- [x] Pass devtools::test() 
- [x] Tested with Exploratory

### DIFF
--- a/R/google_trends.R
+++ b/R/google_trends.R
@@ -106,7 +106,6 @@ getGoogleTrends <- function(user,
     tidyr::gather_(bind_data, "keyword", "trend", colnames(bind_data)[2:ncol(bind_data)], na.rm = TRUE)
   } else if (type == "trend"){
     trend <- ret[[type]]
-    colnames(trend)[colnames(trend) == "hits"] <- "trend"
     trend
   } else {
     stop("Currently, type must be top_regions, top_cities or trends")
--- a/R/google_trends.R
+++ b/R/google_trends.R
@@ -6,7 +6,12 @@
 #' @param last - From when the data should be retreived. Currently supported parameters are "1h", "4h", "1d", "7d", "5y", "30d", "90d", "1y" and "all" (from 2004)
 #' @param geo - Region codes. It's listed in countries data set in gtrendsR.
 #' @export
-getGoogleTrends <- function(user, password, query = "", type = "trend", last = "5y", geo = ""){
+getGoogleTrends <- function(user,
+                            password,
+                            query = "",
+                            type = "trend",
+                            last = "5y",
+                            geo = ""){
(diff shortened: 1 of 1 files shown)


R/google_trends.R
```diff
@@ -77,15 +82,9 @@ getGoogleTrends <- function(user, password, query = "", type = "trend", last = "
...
-    # for output column name instead of the query word.
-    # So we are replacing it with the query word
-    # so that we can use the query word in subsequent steps.
-    if(length(query) == 1){
-      # The query is lowered and spaces are changed into . by gtrendsR, so doing the same
-      colnames(ret[[key]])[[2]] <- stringr::str_to_lower(stringr::str_replace(stringr::str_trim(query), " +", "."))
-    }
+    # Column names should be the same with queries because Japanese character is strangely returned by gtrends package
+    # The query is lowered and spaces are changed into . by gtrendsR, so doing the same
+    colnames(ret[[key]])[-1] <- stringr::str_to_lower(stringr::str_replace(stringr::str_trim(query), " +", "."))
```
hideaki: Can you explain what the negative index [-1] is doing here?

R/google_trends.R
```diff
@@ -106,24 +105,8 @@ getGoogleTrends <- function(user, password, query = "", type = "trend", last = "
...
-    }
-    colnames(trend)[(ncol(trend)-length(cols)+1):ncol(trend)] <- cols
-    # gather columns except for time columns to make it easy to visualise
-    ret <- tidyr::gather_(trend, "keyword", "trend", cols, na.rm = TRUE)
-    if(!is.null(geo)){
-      # separates geo and keywords that were column names
-      ret <- tidyr::separate_(ret, "keyword", c("keyword", "geo"), sep = " ")
-    }
-    ret
+    colnames(trend)[colnames(trend) == "hits"] <- "trend"
```
hideaki: If gtrends itself is changing that much, I think we don't need to try too hard to keep backward compatibility of out output. So I think returning "hits" columns as it is is fine.

R/google_trends.R
double-y: add a comment about it

R/google_trends.R
double-y: Okay, removed it

PR #115, opened 2016-11-28
fix build

double-y:
### Description
install_github has not been working because of namespaces issues (mainly, Import should have been Imports and some loadNamespace should not have been called globally), so fixed them

### Checklist

- [x] Add test cases for this fix/enhancement (in exploratory https://github.com/exploratory-io/tam/pull/3740)
- [x] Pass devtools::check()
- [x] Pass devtools::test() 
- [ ] Tested with Exploratory


NAMESPACE
```diff
@@ -19,6 +19,7 @@ export(createTempEnvironment)
 export(deleteGoogleBigQueryTable)
 export(do_apriori)
 export(do_cmdscale)
+export(do_cmdscale_)
```
double-y: Putting _ is a common way to distinguish NSE function and SE function. dplyr and tidyr also export functions with _ like group_by_, so I also exported it

hideaki: I see!

PR #139, opened 2017-01-10
modify prediction functions

double-y:
### Description
* resolve https://github.com/exploratory-io/tam/issues/4143

* add_prediction for unknown categories support

* filter data only when prediction causes error because it can work if there is only one categorical variable

### Checklist

- [x] Pass devtools::check()
- [x] Pass devtools::test()
- [x] Tested with Exploratory

### DIFF
--- a/R/broom_wrapper.R
+++ b/R/broom_wrapper.R
@@ -94,8 +94,33 @@ predict <- function(df, model, ...){
 
 #' apply data frame with model to a data frame
 #' @export
-add_prediction <- function(df, model_df){
-  broom::augment(model_df, model, newdata = df)
+add_prediction <- function(df, model_df, ...){
+
+  tryCatch({
+    broom::augment(model_df, model, newdata = df, ...)
+  }, error = function(e){
+    if (grepl("arguments imply differing number of rows: ", e$message)) {
+      # in this case, df has categories that aren't in model
+
+      filtered_data <- df
+
+      for(model in model_df[["model"]]){
+        # remove rows that have categories that aren't in model
+        # otherwise, broom::augment causes an error
+        for (cname in colnames(model$model)) {
+          filtered_data <- filtered_data[filtered_data[[cname]] %in% model$model[[cname]], ]
+        }
+      }
+
+      if (nrow(filtered_data) == 0) {
+        stop("not enough information to predict in data frame")
+      }
+
+      broom::augment(model_df, model, newdata = filtered_data, ...)
+    } else {
+      stop(e$message)
+    }
+  })
 }
 
 #' assign cluster number to each rows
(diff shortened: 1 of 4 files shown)


R/broom_wrapper.R
```diff
@@ -205,31 +230,61 @@ prediction <- function(df, source_data, test = TRUE){
...
   df <- dplyr::select(df, .test_index, model)
+  cll <- match.call()
+  aug_args <- expand_args(cll, exclude = c("df", "source_data", "test"))
 
   ret <- if(test){
     # augment by test data
-    dplyr::bind_cols(df, source) %>%
+    # use formula to support expanded aug_args (especially for type.predict for logistic regression)
+    aug_fml <- if(aug_args == ""){
+      as.formula(paste0("~list(broom::augment(model, newdata = data))"))
```
hideaki: Is the paste0() here necessary?

R/broom_wrapper.R
```diff
@@ -205,31 +230,61 @@ prediction <- function(df, source_data, test = TRUE){
 
   # drop unnecessary columns
   df <- dplyr::select(df, .test_index, model)
+  cll <- match.call()
```
hideaki: Please add comment on what these lines are doing.

R/broom_wrapper.R
```diff
@@ -185,7 +210,7 @@ kmeans_info <- function(df){
 #' @param source_data Data frame that has model and .test_index
```
hideaki: The @param comment for source_data is same as df. Is this correct?

R/broom_wrapper.R
```diff
@@ -205,31 +230,61 @@ prediction <- function(df, source_data, test = TRUE){
...
+      }
+    })
+
+    ret <- augmented %>%
       dplyr::select(-model) %>%
       tidyr::unnest(data)
+
   } else {
     # augment by trainig data
+    # use formula to support expanded aug_args (especially for type.predict for logistic regression)
```
hideaki: This part looks same as the case test is TRUE. Can it be outside of if statement to avoid duplicate?

R/broom_wrapper.R
```diff
@@ -94,8 +94,33 @@ predict <- function(df, model, ...){
...
+
+  tryCatch({
+    broom::augment(model_df, model, newdata = df, ...)
+  }, error = function(e){
+    if (grepl("arguments imply differing number of rows: ", e$message)) {
+      # in this case, df has categories that aren't in model
+
+      filtered_data <- df
+
+      for(model in model_df[["model"]]){
```
hideaki: Looks like this is looping over all the columns, but doesn't this do the filtering on numeric columns too?

R/broom_wrapper.R
```diff
@@ -94,8 +94,33 @@ predict <- function(df, model, ...){
...
 #' @export
-add_prediction <- function(df, model_df){
-  broom::augment(model_df, model, newdata = df)
+add_prediction <- function(df, model_df, ...){
+
+  tryCatch({
+    broom::augment(model_df, model, newdata = df, ...)
+  }, error = function(e){
+    if (grepl("arguments imply differing number of rows: ", e$message)) {
+      # in this case, df has categories that aren't in model
```
hideaki: This deserves more explanation.

R/broom_wrapper.R
```diff
@@ -205,31 +230,61 @@ prediction <- function(df, source_data, test = TRUE){
 
   # drop unnecessary columns
   df <- dplyr::select(df, .test_index, model)
+  cll <- match.call()
+  aug_args <- expand_args(cll, exclude = c("df", "source_data", "test"))
 
   ret <- if(test){
     # augment by test data
-    dplyr::bind_cols(df, source) %>%
+    # use formula to support expanded aug_args (especially for type.predict for logistic regression)
```
hideaki: Creating formula seems tricky, but what was the reason you had to do this?

R/broom_wrapper.R
```diff
@@ -205,31 +230,61 @@ prediction <- function(df, source_data, test = TRUE){
...
-      dplyr::select(-.test_index) %>%
-      dplyr::rowwise() %>%
-      dplyr::mutate(data = list(broom::augment(model, newdata = data))) %>%
+      dplyr::select(-.test_index)
+
+
+    augmented <- tryCatch({
+      data_to_augment %>%
+        dplyr::rowwise() %>%
+        dplyr::mutate_(.dots = list( data = aug_fml))
```
hideaki: This line looks like black magic to me. Some comment would help.

R/broom_wrapper.R
double-y: Will add it

R/broom_wrapper.R
double-y: Right, will add a condition to check if they are numeric

R/broom_wrapper.R
double-y: No. Will modify it

R/broom_wrapper.R
double-y: Will add it

R/broom_wrapper.R
double-y: Will add a comment for it

R/broom_wrapper.R
double-y: No, will remove it

R/broom_wrapper.R
double-y: Do you mean aug_fml? It uses newdata for the argument if test is TRUE. Otherwise, it uses data for the argument

hideaki: I see. Can you add a comment about it? Just by looking, it's hard to see what is the difference between them.

double-y: Okay, will add a comment

PR #141, opened 2017-01-11
add scale wrapper

double-y:
### Description

Add a scale wrapper function

Related to https://github.com/exploratory-io/tam/issues/3991

### Checklist

- [x] Pass devtools::check()
- [x] Pass devtools::test()
- [x] Tested with Exploratory


R/stats_wrapper.R
```diff
@@ -1,6 +1,13 @@
 #'
 #'
 
+#' scale wrapper that returns a vector as a result
+#' @export
+scale_vec <- function(...) {
+  ret <- scale(...)
+  as.numeric(ret)
```
hideaki: Not very sure what is going on here.
Looks like only difference with regular scale() is as.numeric().
Does this mean as.numeric have side effect of making non-vector a vector?
Please explain a little more in the comment.

double-y: I see. I'll explain it there.

PR #142, opened 2017-01-12
update prediction functions

double-y:
### Description
prediction functions return both link and response if type.predict is not indicated and the models are glm

### Checklist

- [x] Pass devtools::check()
- [x] Pass devtools::test()
- [x] Tested with Exploratory

### DIFF
--- a/R/broom_wrapper.R
+++ b/R/broom_wrapper.R
@@ -98,9 +98,42 @@ predict <- function(df, model, ...){
 #' @param ... Additional argument to be passed to broom::augment
 #' @export
 add_prediction <- function(df, model_df, ...){
+  # parsing arguments of add_prediction and getting optional arguemnt for augment in ...
+  cll <- match.call()
+  aug_args <- expand_args(cll, exclude = c("df", "model_df"))
 
-  tryCatch({
-    broom::augment(model_df, model, newdata = df, ...)
+  get_result_with_response <- function(model_df, df, aug_args){
+    # Use formula to support expanded aug_args (especially for type.predict for logistic regression)
+    # because ... can't be passed to a function inside mutate directly.
+    aug_fml <- if(aug_args == ""){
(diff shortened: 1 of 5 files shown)


tests/testthat/test_build_glm.R
```diff
@@ -118,12 +118,25 @@ test_that("prediction with categorical columns", {
...
 
   ret <- prediction(model_data, test_data, type.predict = "response")
+  both_ret <- prediction(model_data, test_data)
+  train_ret <- prediction(model_data, test_data, test = FALSE)
+
   expect_true(nrow(ret) > 0)
   expect_true(all(ret["Fitted"] >= 0 & ret["Fitted"] <= 1))
   expect_equal(colnames(ret), c("CANCELLED", "Carrier.Name", "CARRIER", "DISTANCE", "Fitted", "Standard Error"))
 
+  expect_true(all(both_ret["Fitted.response"] >= 0 & both_ret["Fitted.response"] <= 1))
```
hideaki: Why does prediction adds "Fitted.response" column while add_prediction adds "fitted.response"?

double-y: Fitted.response appears in model info view but fitted.response appears in data frame, so I think lower case is better for it

R/broom_wrapper.R
```diff
@@ -289,6 +341,13 @@ prediction <- function(df, source_data, test = TRUE, ...){
       }
     })
 
+    if (with_response){
+      augmented <- augmented %>%
+        dplyr::ungroup() %>%
+        dplyr::mutate(data = purrr::map2(data, model, add_response)) %>%
+        dplyr::rowwise()
```
hideaki: What is this rowwise() for?

double-y: Ah, probably not needed

PR #146, opened 2017-01-14
update model functions

double-y:
### Description
- [x] remove resetting grouping logic from build_lm and build_glm
- [x] prediction function uses source.data column, not given source data frame
- [x] prediction function should have an argument to clean column names of result
- [x] add do_cor as function for ROC curve
- [x] add build_lr for logistic regression
- [x] add evaluation functions for regression and binary classification
- [x] https://github.com/exploratory-io/tam/issues/4279
- [x] fix https://github.com/exploratory-io/tam/issues/4280 (grouping error message handling)

### Checklist

- [ ] Pass devtools::check()
- [ ] Pass devtools::test()
- [ ] Tested with Exploratory


R/build_lm.R
```diff
@@ -67,40 +63,35 @@ build_lm <- function(data, ..., keep.source = TRUE, augment = FALSE, group_cols
...
           eval(parse(text = paste0("stats::lm(data = data, ", arg_char, ")")))
         })) %>%
-        dplyr::select(-data) %>%
         dplyr::rowwise()
       class(ret[[source_col]]) <- c("list", ".source.data")
       ret
     } else {
       ret <- data %>%
-        tidyr::nest() %>%
+        tidyr::nest(.key = "source.data") %>%
```
hideaki: line 83 to 93 looks same as line 66 to 77. Can we avoid code duplication here?

R/model_eval.R
```diff
@@ -0,0 +1,264 @@
...
+    val <- arranged[[actual_val_col]][!is.na(arranged[[actual_val_col]])]
+    if(is.factor(val)){
+      # Need to subtract 1 because the first level in factor is regarded as 1
+      # though it should be FALSE.
+      val <- as.logical(as.integer(val) - 1)
+    } else {
+      val <- as.logical(val)
+    }
+
+    if (all(val)){
```
hideaki: Please add some more comment below this point of this function.

R/build_lm.R
double-y: Yes, it's code duplication. Removed it

R/model_eval.R
double-y: Okay, will add comments

PR #151, opened 2017-01-19
modify output column name

double-y:
### Description
output column name change for model functions

### Checklist

- [x] Pass devtools::check()
- [x] Pass devtools::test()
- [x] Tested with Exploratory


R/broom_wrapper.R
```diff
@@ -168,13 +168,13 @@ add_prediction <- function(df, model_df, ...){
     }
   })
   # update column name based on both link and response are there for fitted values
-  fitted_label <- if("fitted_response" %in% colnames(ret)){
-    "fitted_link"
+  fitted_label <- if("predicted_response" %in% colnames(ret)){
+    "predicted_link"
```
hideaki: What is the case we use the name "predicted_link"?

PR #153, opened 2017-01-20
update model funcs

double-y:
### Description
* modify an error message

* fix a test case for conf_int

### Checklist

- [x] Pass devtools::check()
- [x] Pass devtools::test()
- [ ] Tested with Exploratory

### DIFF
--- a/R/build_glm.R
+++ b/R/build_glm.R
@@ -88,7 +88,7 @@ build_glm <- function(data, formula, ..., keep.source = TRUE, augment = FALSE, g
     ret
   }, error = function(e){
     if(e$message == "contrasts can be applied only to factors with 2 or more levels"){
-      stop("more than 1 unique value is expected for categorical columns assigned as predictors")
+      stop("more than 1 unique values are expected for categorical columns assigned as predictors")
     }
 
     # cases when a grouping column is in variables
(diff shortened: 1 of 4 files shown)


R/build_glm.R
```diff
@@ -88,7 +88,7 @@ build_glm <- function(data, formula, ..., keep.source = TRUE, augment = FALSE, g
     ret
   }, error = function(e){
     if(e$message == "contrasts can be applied only to factors with 2 or more levels"){
-      stop("more than 2 unique values are needed for categorical predictor columns")
+      stop("more than 1 unique value is expected for categorical columns assigned as predictors")
```
hideaki: Should be , "more than 1 unique values are expected..."

R/build_lm.R
```diff
@@ -85,7 +85,7 @@ build_lm <- function(data, ..., keep.source = TRUE, augment = FALSE, group_cols
     ret
   }, error = function(e){
     if(e$message == "contrasts can be applied only to factors with 2 or more levels"){
-      stop("more than 2 unique values are needed for categorical predictor columns")
+      stop("more than 1 unique value is expected for categorical columns assigned as predictors")
```
hideaki: Should be , "more than 1 unique values are expected..."

R/build_glm.R
double-y: Right, will modify so

R/build_lm.R
double-y: Right, will modify so

PR #154, opened 2017-01-20
modify order of arguments

double-y:
### Description
Use predicted column name as first argument and actual value column as second argument to fit it to command UI

### Checklist

- [x] Pass devtools::check()
- [x] Pass devtools::test()
- [x] Tested with Exploratory

### DIFF
--- a/R/model_eval.R
+++ b/R/model_eval.R
@@ -1,20 +1,20 @@
 #' Non standard evaluation version of do_roc_
 #' @param df Data frame
-#' @param actual_val Column name for actual values
 #' @param pred_prob Column name for probability
+#' @param actual_val Column name for actual values
 #' @export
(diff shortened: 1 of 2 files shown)


R/model_eval.R
```diff
@@ -145,21 +145,21 @@ evaluate_binary_ <- function(df, actual_val_col, pred_prob_col, threshold = "f_s
 
 #' Non standard evaluation version of evaluate_regression_
 #' @param df Data frame
-#' @param actual_val Column name for actual values
 #' @param fitted Column name for predicted values
```
hideaki: Should fitted renamed to pred_val? That sounds more in line with the names we use for output columns.

double-y: I see. Okay, will modify so

PR #155, opened 2017-01-28
Changed to pass encoding argument to make it work with Japanese environment.

hidekoji:
### Description

This PR addresses the issue that exploratory::checkSourceConflict does not work with Japanese file.

### Checklist

- [ ] Pass devtools::check()
- [ ] Pass devtools::test()
- [ ] Tested with Exploratory


R/system.R
```diff
@@ -1057,7 +1057,7 @@ checkSourceConflict <- function(files){
   for (file in files){
     ret[[file]] <- tryCatch({
       env <- new.env()
-      source(file, local=env)
+      source(file, local=env, encoding = "UTF-8")
```
hideaki: Please add some comment on why it is safe to assume the file is always in UTF-8.

PR #184, opened 2017-02-13
modify eval func for labels with 2 unique values

double-y:
### Description
Support actual labels with 2 unique values

Fix https://github.com/exploratory-io/tam/issues/4679

### Checklist

- [x] Pass devtools::check()
- [x] Pass devtools::test()
- [ ] Tested with Exploratory


R/util.R
```diff
@@ -705,3 +705,26 @@ pivot_ <- function(data, formula, value_col = NULL, fun.aggregate = mean, fill =
...
+#'convert values to binary label
+binary_label <- function(val) {
+  if(is.factor(val)){
+    # Need to subtract 1 because the first level in factor is regarded as 1
+    # though it should be FALSE.
+    val <- as.logical(as.integer(val) - 1)
+  } else {
+    logi_val <- as.logical(val)
+    # if the values are non-zero values,
+    # larger value should be regarrded as FALSE
```
hideaki: typo regarrded

R/util.R
```diff
@@ -705,3 +705,26 @@ pivot_ <- function(data, formula, value_col = NULL, fun.aggregate = mean, fill =
...
+    if(all(logi_val[!is.na(logi_val)])){
```
hideaki: some more comments on the logic around here would help.

R/util.R
```diff
@@ -705,3 +705,26 @@ pivot_ <- function(data, formula, value_col = NULL, fun.aggregate = mean, fill =
...
+      unique_val <- unique(val[!is.na(val)])
+      if(length(unique_val) == 2) {
+        logi_val <- val == max(unique_val)
+      } else if (length(unique_val) > 2) {
+        stop("actual labels can't have more than 2 unique values")
```
hideaki: To be a generic utility function, we should not assume that the column is called "actual".

double-y: modified to "binary labels can't have more than 2 unique values"

R/util.R
double-y: Added comments

R/util.R
double-y: Fixed it

PR #186, opened 2017-02-13
time-dependent AUC of survival model

hideaki:
### Description
Added do_survauc to calculate time-dependent AUC of survival model.

### Checklist

- [ ] Pass devtools::check()
- [x] Pass devtools::test() - no extra errors
- [ ] Tested with Exploratory


hideaki:
We will probably use other package for this, since methods like "uno" is giving result that is hard to explain.

PR #196, opened 2017-02-23
update xgboost functions

double-y:
### Description
Made xgboost_reg, xgboost_binary, xgboost_multi for build_model and separated tidier methods for them

resolve https://github.com/exploratory-io/tam/issues/4752

### Checklist

- [x] Pass devtools::check()
- [x] Pass devtools::test()
- [x] Tested with Exploratory


R/util.R
```diff
@@ -265,7 +265,13 @@ mat_to_df <- function(mat, cnames=NULL, na.rm=TRUE, diag=TRUE){
 #' match the type of two vector
 same_type <- function(vector, original){
   if(is.factor(original)){
-    as.factor(vector)
+    if(all(vector[!is.na(vector)] %in% levels(original))){
+      # if original is factor and vector has all values,
+      # should return factor with same levels
+      factor(vector, levels = vector)
```
hideaki: Did you mean something like this?
factor(vector, levels = levels(original))

R/build_xgboost.R
```diff
@@ -96,53 +20,144 @@ fml_xgboost <- function(data, formula, weight = NULL, params = list(), ...) {
...
+  } else {
+    factored <- as.factor(data[[y_name]])
+    y_vals <- as.numeric(factored) - 1
+    same_type(levels(factored), data[[y_name]])
+  }
+
+  data[[y_name]] <- y_vals
+
+  objective <- paste0("multi:", output_type, sep = "")
+  ret <- ret <- data_xgboost(data = data,
```
hideaki: why is ret substituted twice?

double-y: It's my mistake. Will remove it

R/util.R
double-y: Yes, it's mistake. Added test case and fixed

PR #201, opened 2017-02-26
fix wrong data size values

double-y:
### Description
data size should have been numbers of each class but it was the number of whole data

### Checklist

- [x] Pass devtools::check()
- [x] Pass devtools::test()
- [x] Tested with Exploratory


R/randomForest_tidiers.R
```diff
@@ -1,14 +1,152 @@
+#' Random Forest wrapper for regression
```
hideaki: Please add comment about the following 2 differences from raw randomForest as a comment for this function as a whole.
- When . is used in right hand side of formula, only numeric/logical columns are used as predictors.
- terms_mapping attribute is added to model for keeping mapping of original column names and cleaned-up column names.

PR #208, opened 2017-02-27
Make cohort names shorter by omitting columns with only 1 unique value

hideaki:
### Description
Make cohort names shorter by omitting columns with only 1 unique value
As a column to distinguish cohort, we create column like age_sex_ph.ecog,
but for example, if there is only 1 type of sex in newdata, it will be age_ph.ecog.

### Checklist

- [ ] Pass devtools::check()
- [x] Pass devtools::test()
- [ ] Tested with Exploratory

### DIFF
--- a/R/broom_wrapper.R
+++ b/R/broom_wrapper.R
@@ -848,6 +848,20 @@ model_anova <- function(df, pretty.name = FALSE){
   ret
 }
 
+# function to use in prediction_survfit to remove single value columns.
+# alternative to select_if which gives error when column name has '-' in it.
+non_single_value_colnames <- function(data) {
+  ret <- c()
+  for (colname in colnames(data)) {
+    if (length(unique(data[[colname]])) > 1) {
+      ret <- c(ret, colname)
+    } else if (is.logical(data[[colname]])) {
+      ret <- c(ret,colname)
+    }
+  }
+  ret
+}
+
 #' tidy after converting model to survfit
 #' @param newdata Data frame with rows that represent cohorts to simulate
 #' @export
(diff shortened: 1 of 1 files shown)


R/broom_wrapper.R
```diff
@@ -879,12 +879,14 @@ prediction_survfit <- function(df, newdata = NULL, ...){
     # convert characterized data back to numeric.
     ret <- ret %>% mutate(estimate = as.numeric(estimate), std_error = as.numeric(std_error), conf_high = as.numeric(conf_high), conf_low = as.numeric(conf_low))
     # replace the cohort name with a string that is a concatenation of values that represents the cohort.
-    cohorts_labels <- newdata %>% unite(label, everything())
-    for (i in 1:nrow(newdata)){
+    # but, omit newdata column that has only 1 unique value from the cohort names we create here.
+    selected_newdata <- newdata %>% dplyr::select_if(function(x) length(unique(x)) > 1)
```
double-y: I found that select_if throws an error if the column names are not clean. Is it okay?

R/broom_wrapper.R
```diff
@@ -879,12 +893,15 @@ prediction_survfit <- function(df, newdata = NULL, ...){
     # convert characterized data back to numeric.
     ret <- ret %>% mutate(estimate = as.numeric(estimate), std_error = as.numeric(std_error), conf_high = as.numeric(conf_high), conf_low = as.numeric(conf_low))
     # replace the cohort name with a string that is a concatenation of values that represents the cohort.
-    cohorts_labels <- newdata %>% unite(label, everything())
-    for (i in 1:nrow(newdata)){
+    # but, omit newdata column that has only 1 unique value from the cohort names we create here.
+    selected_newdata_colnames <- non_single_value_colnames(newdata)
+    selected_newdata <- newdata[, selected_newdata_colnames]
```
double-y: Can this be length 0? If so, 1:nrow(selected_newdata) in l900 might cause an error

PR #212, opened 2017-03-02
fix kmeans to respect group_by

double-y:
### Description
fix kmeans to respect group_by

### Checklist

- [x] Pass devtools::check()
- [x] Pass devtools::test()
- [x] Tested with Exploratory

### DIFF
--- a/R/broom_wrapper.R
+++ b/R/broom_wrapper.R
@@ -42,7 +42,7 @@ augment_kmeans <- function(df, model, data){
     augment_func <- get("augment", asNamespace("broom"))
     ret <- do.call(augment_func, list(df, model_col, data=data_col))
     # cluster column is factor labeled "1", "2"..., so convert it to integer to avoid confusion
-    ret[[ncol(ret)]] <- as.integer(ret[[ncol(ret)]])
+    ret[[ncol(ret)]] <- ret[[ncol(ret)]]
     ret
   },
   error = function(e){
--- a/R/model_builder.R
+++ b/R/model_builder.R
@@ -96,8 +96,8 @@ build_kmeans.kv_ <- function(df,
     )
     if(augment){
       cluster_column <- avoid_conflict(grouped_column, "cluster")
-      row_fact <- factor(df[[row_col]], levels=seq(centers))
-      df[[cluster_column]] <- as.factor(kmeans_ret$cluster[row_fact])
+      row_fact <- as.factor(df[[row_col]])
+      df[[cluster_column]] <- factor(kmeans_ret$cluster[row_fact], levels = seq(centers))
       df
     } else {
       # add an attribute to be referred from augment_kmeans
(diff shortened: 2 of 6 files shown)


R/broom_wrapper.R
```diff
@@ -42,7 +42,7 @@ augment_kmeans <- function(df, model, data){
     augment_func <- get("augment", asNamespace("broom"))
     ret <- do.call(augment_func, list(df, model_col, data=data_col))
     # cluster column is factor labeled "1", "2"..., so convert it to integer to avoid confusion
```
hideaki: We keep it as factor. So this comment should be modified to reflect that.

double-y: fixed it

R/model_builder.R
```diff
@@ -201,8 +192,8 @@ build_kmeans.cols <- function(df, ...,
     if(augment){
       ret <- broom::augment(kmeans_ret, df)
       colnames(ret)[[ncol(ret)]] <- avoid_conflict(grouped_column, "cluster")
-      # cluster column is factor labeled "1", "2"..., so convert it to integer to avoid confusion
-      ret[[ncol(ret)]] <- as.integer(ret[[ncol(ret)]])
+      # make the clustering labels factor with levels with sequence of numbers
+      ret[[ncol(ret)]] <- factor(ret[[ncol(ret)]], levels=seq(centers))
```
hideaki: Isn't ret[[ncol(ret)]] already a factor according to the code before change?
If so do we need to apply factor() again?

R/model_builder.R
```diff
@@ -169,13 +158,15 @@ build_kmeans.cols <- function(df, ...,
   }
 
   build_kmeans_each <- function(df){
+    # remove grouping column
```
hideaki: why do we need to remove grouping column?

R/model_builder.R
```diff
@@ -102,7 +97,7 @@ build_kmeans.kv_ <- function(df,
     if(augment){
       cluster_column <- avoid_conflict(grouped_column, "cluster")
       row_fact <- as.factor(df[[row_col]])
-      df[[cluster_column]] <- kmeans_ret$cluster[row_fact]
+      df[[cluster_column]] <- factor(kmeans_ret$cluster[row_fact], levels = seq(centers))
```
hideaki: Could you add comment on what this line is doing?

R/model_builder.R
double-y: Right, will remove it

R/model_builder.R
```diff
@@ -77,6 +70,8 @@ build_kmeans.kv_ <- function(df,
   }
 
   build_kmeans_each <- function(df){
+    # remove grouping column
```
hideaki: why do we need to remove grouping column here?

R/model_builder.R
double-y: I'll add comments

R/model_builder.R
double-y: Added comments

R/model_builder.R
double-y: Added a comment

PR #219, opened 2017-03-11
change predict_na to impute_na

double-y:
### Description
change predict_na to impute_na and it supports mean and median to fill

### Checklist

- [x] Pass devtools::check()
- [x] Pass devtools::test()
- [ ] Tested with Exploratory


R/na_util.R
```diff
@@ -1,15 +1,28 @@
...
-    # predict where target is NA
-    predict(lm_model, df[is.na(target),])
+impute_na <- function(target, ..., type = "lm_predict") {
+  ret <- switch(type, lm_predict = {
+    # list(...) is a list of vectors
+    df <- as.data.frame(list(...))
+    df$target <- target
+    lm_model <- lm(data = df, target ~ ., na.action = na.omit)
+    # predict return double, so values should be double
+    ret <- as.double(target)
```
hideaki: Shouldn't we convert lm return to target type rather than converting target to double?

PR #220, opened 2017-03-11
add cluster function

double-y:
### Description
add cluster function

resolve https://github.com/exploratory-io/tam/issues/5058

### Checklist

- [x] Pass devtools::check()
- [x] Pass devtools::test()
- [ ] Tested with Exploratory


R/cluster.R
```diff
@@ -0,0 +1,30 @@
+#' cluster based on vectors
```
hideaki: Please add a little more explanation on what this function is as a whole.

R/cluster.R
```diff
@@ -0,0 +1,30 @@
...
+#' @export
+cluster <- function(..., type = "kmeans", n_cluster = 3) {
+  data <- list(...) %>% as.data.frame() %>% as.matrix()
+  na_removed <- data %>% na.omit()
+  if(nrow(na_removed) == 0) {
```
hideaki: Does k-means work for example when nrow() is 1 or below n_cluster?
If not, we can limit this condition further.

R/cluster.R
```diff
@@ -0,0 +1,30 @@
...
+    stop("no valid data to cluster")
```
hideaki: This message is a little too terse to understand what it is quickly. We could be a little more explanatory like this.
"not enough valid data to perform clustering"

R/cluster.R
```diff
@@ -0,0 +1,30 @@
...
+  }
+  model <- kmeans(na_removed, centers = n_cluster)
+  cluster <- if (ncol(data) == 1) {
```
hideaki: Please add comment on why we do this distinction on ncol size.

R/cluster.R
double-y: added comments

R/cluster.R
double-y: added a condition

R/cluster.R
double-y: modified the message

R/cluster.R
double-y: added comments

PR #221, opened 2017-03-11
upgrade xgboost functions

double-y:
### Description
* support non-numeric predictor columns
* support validation data split for watchlist
* return all metrics

### Checklist

- [ ] Pass devtools::check()
- [ ] Pass devtools::test()
- [ ] Tested with Exploratory


R/test_wrapper.R
```diff
@@ -102,6 +102,54 @@ do_var.test <- function(df, value, key, ...){
   df %>% dplyr::do_(.dots=setNames(list(~do_var.test_each(df = ., ...)), model_col)) %>% tidyr::unnest_(model_col)
 }
 
+#' chisq.test wrapper
```
hideaki: Shouldn't we remove old do_chisq.test_?

R/build_xgboost.R
```diff
@@ -1,58 +1,70 @@
...
-  if(is.integer(x) || is.logical(x)){
-    # x must be a double matrix
-    x <- matrix(as.double(x), ncol = ncol(x))
-  }
-
-  if (nrow(x) == 0) {
+#' @export
+fml_xgboost <- function(data, formula, nrounds= 10, weights = NULL, watchlist_rate = 0, ...) {
+  term <- terms(formula, data = data)
+  # do.call is used to substitute weight
```
hideaki: Nit picking, but it's weights, not weight.

R/build_xgboost.R
```diff
@@ -1,58 +1,70 @@
...
+
+    y <- model.response(md_frame)
+
+
+    ret <- if(watchlist_rate != 0.0) {
+      if (watchlist_rate < 0 ||  1 <= watchlist_rate) {
+        stop("watchlist_rate must be between 0 and 1")
+      }
+      # create validation data set
+      index <- sample(seq(nrow(md_mat)), ceiling(nrow(md_mat) * watchlist_rate))
```
hideaki: Looks like this validation data split is just base on nrow, but shouldn't we randomize it?
Also, wasn't xgboost doing this split internally just by giving ratio? What is the reason you chose to do the split ourselves?

R/build_xgboost.R
```diff
@@ -331,17 +406,25 @@ tidy.xgb.Booster <- function(x, type="weight", pretty.name = FALSE, ...){
...
+      colnames(ret)[colnames(ret)=="RealCover"] <- "Real Coverage"
+      colnames(ret)[colnames(ret)=="RealCover %"] <- "Real Coverage %"
     } else {
       colnames(ret)[colnames(ret)=="Feature"] <- "feature"
       colnames(ret)[colnames(ret)=="Gain"] <- "importance"
       colnames(ret)[colnames(ret)=="Cover"] <- "coverage"
       colnames(ret)[colnames(ret)=="Frequency"] <- "frequency"
+      colnames(ret)[colnames(ret)=="Split"] <- "split"
+      colnames(ret)[colnames(ret)=="RealCover"] <- "real_coverage"
+      colnames(ret)[colnames(ret)=="RealCover %"] <- "real_coverage_%"
```
hideaki: I think we are using "pct" to express percent in other places.

R/util.R
```diff
@@ -828,3 +836,20 @@ get_multi_predicted_values <- function(prob_mat, actual_vals = NULL){
   ret$predicted_label <- prob_label
   ret
 }
+
+fill_vec_NA <- function(indice, values, max_index = max(indice, na.rm = TRUE)){
```
hideaki: Please add some comment on what this function is.

R/util.R
```diff
@@ -828,3 +836,20 @@ get_multi_predicted_values <- function(prob_mat, actual_vals = NULL){
...
+  ret <- same_type(rep(NA, max_index), values)
+  ret[indice] <- values
+  ret
+}
+
+fill_mat_NA <- function(indice, mat, max_index = max(indice, na.rm = TRUE)){
```
hideaki: Please add some comment on what this function is.

R/test_wrapper.R
double-y: I mistakingly created duplicated do_chisq.test_. Will remove it

R/util.R
double-y: added comments

R/util.R
double-y: added comments

R/build_xgboost.R
double-y: sample function does randomize
```r
> sample(10, ceiling(10 * 0.3))
[1] 3 9 7
```

xgboost just takes matrix and labels set for validation data, so it has to be created beforehand

R/build_xgboost.R
double-y: I see. modified so

PR #225, opened 2017-03-17
modify impute_na function

double-y:
### Description
* add empty predictor selection error handling
* default type is mean
* type should support 'value' and aggregate function

fix https://github.com/exploratory-io/tam/issues/5138

### Checklist

- [x] Pass devtools::check()
- [x] Pass devtools::test()
- [ ] Tested with Exploratory

### DIFF
--- a/R/na_util.R
+++ b/R/na_util.R
@@ -2,7 +2,7 @@
 #' @param target Target vector whose NA is filled
 #' @param ... Vectors to be used to pridict NA when type is lm_predict
 #' @param type This can be "mean", "median", "lm_predict", "value" or aggregate function
-#' @paramm val This is effective when type is mean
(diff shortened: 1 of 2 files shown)


R/na_util.R
```diff
@@ -1,30 +1,52 @@
 #' Guess missing values by lm
+#' @param target Target vector whose NA is filled
+#' @param ... Vectors to be used to pridict NA when type is lm_predict
+#' @param type This can be "mean", "median", "lm_predict", "value" or aggregate function
+#' @paramm val This is effective when type is mean
```
hideaki: Isn't val effective when type is "value"?

double-y: Right. Modified

PR #229, opened 2017-03-22
add column type validation for prediction

double-y:
### Description
* add column type validation for newdata prediction (fix https://github.com/exploratory-io/tam/issues/5145 )
* add missing na.action argument to randomForest functions

### Checklist

- [x] Pass devtools::check()
- [x] Pass devtools::test()
- [x] Tested with Exploratory

### DIFF
--- a/R/broom_wrapper.R
+++ b/R/broom_wrapper.R
@@ -88,7 +88,7 @@ add_prediction <- function(df, model_df, conf_int = 0.95, ...){
   aug_args <- expand_args(cll, exclude = c("df", "model_df"))
 
   # validate data frame based on meta info
-  model_meta <- model_df[[".model_meta_data"]]
+  model_meta <- model_df[[".model_metadata"]]
   if(!is.null(model_meta)){
     types <- model_meta[[1]]$types
     if(!is.null(types)){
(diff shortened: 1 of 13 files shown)


R/util.R
```diff
@@ -914,3 +914,54 @@ add_confint <- function(data, conf_int){
...
+
+    if(any(!is.na(message))){
+      stop(paste0("Data type mismatch. ", paste0(message[!is.na(message)], collapse = ", ")))
+    }
+  }
+  TRUE
+}
+
+# used in model building functions to create meta data
+# the meta data is used for column type validation
```
hideaki: some more explanation on the purpose and content of this metadata would help.

double-y: Added comments

R/broom_wrapper.R
```diff
@@ -88,7 +88,7 @@ add_prediction <- function(df, model_df, conf_int = 0.95, ...){
   aug_args <- expand_args(cll, exclude = c("df", "model_df"))
 
   # validate data frame based on meta info
-  model_meta <- model_df[[".model_meta_information"]]
+  model_meta <- model_df[[".model_meta_data"]]
```
hideaki: Sorry for being picky on the name, but I think it's usually "metadata" than "meta data".

R/util.R
```diff
@@ -947,8 +947,11 @@ validate_data <- function(types, data){
   TRUE
 }
 
-# used in model building functions to create meta data
-# the meta data is used for column type validation
+# This is used in model building functions
+# to create meta data.
+# It contains terms to create model and
+# column types of the variables.
+# Its's used for column type validation.
```
hideaki: Example of data structure would help understanding what this is doing.
Based on our discussion today, comment like following would help.

# return value looks like following example.
# list(type = list(col1 = "numeric", col2 = "character"))

I think this is important enough to be explained well since it becomes basis of how we deal with model.

double-y: Ok, added a comment

R/broom_wrapper.R
double-y: I see. Okay, will modify so

PR #239, opened 2017-03-30
add copied token

double-y:
### Description
Add an R6 object for Oauth2.0 token wrapper

### Checklist

- [x] Pass devtools::check()
- [ ] Pass devtools::test()
- [x] Tested with Exploratory


R/oauth_wrapper.R
```diff
@@ -0,0 +1,24 @@
+#' @export
+CopiedToken2.0 <- R6::R6Class("CopiedToken2.0", inherit = httr::Token2.0, list(
```
hideaki: When in exploratory package's context, I think this should be called something like HttrOAuthToken2.0, since being about httr's OAuth is not an obvious thing here.

double-y: renamed it

PR #240, opened 2017-03-31
update google analytics for server

double-y:
### Description
update google analytics for server

### Checklist

- [ ] Pass devtools::check()
- [ ] Pass devtools::test()
- [ ] Tested with Exploratory


R/oauth.R
```diff
@@ -60,11 +60,32 @@ getGoogleTokenForAnalytics <- function(tokenFileId = "", useCache=TRUE){
...
+      validate = "https://www.googleapis.com/oauth2/v1/tokeninfo",
+      revoke = "https://accounts.google.com/o/oauth2/revoke",
+      appname = appName,
+      key = clientId,
+      secret = secret,
+      credentials = list(
+        access_token = token_info$access_token,
+        refresh_token = token_info$token_type,
+        token_type = token_info$token_type,
+        expiry_date = token_info$expiry_data
```
hideaki: should be $expiry_date instead of $expiry_data

R/oauth.R
```diff
@@ -60,11 +60,32 @@ getGoogleTokenForAnalytics <- function(tokenFileId = "", useCache=TRUE){
...
+      authorize = "https://accounts.google.com/o/oauth2/auth",
+      access = "https://accounts.google.com/o/oauth2/token",
+      validate = "https://www.googleapis.com/oauth2/v1/tokeninfo",
+      revoke = "https://accounts.google.com/o/oauth2/revoke",
+      appname = appName,
+      key = clientId,
+      secret = secret,
+      credentials = list(
+        access_token = token_info$access_token,
+        refresh_token = token_info$token_type,
```
hideaki: Should be $refresh_token
If this worked, maybe we didn't need refresh_token here in the first place?

R/oauth.R
```diff
@@ -60,11 +60,32 @@ getGoogleTokenForAnalytics <- function(tokenFileId = "", useCache=TRUE){
...
+        token_type = token_info$token_type,
```
hideaki: Shouldn't this always be "Bearer"? If so we can hard code.

R/oauth.R
double-y: Hmm, it might be true but when I was testing Validation, it was needed. I think having it is safer

R/oauth.R
double-y: Right, will fix it

PR #254, opened 2017-04-09
update app data functions

double-y:
### Description
Update functions to call api

### Checklist

- [x] Pass devtools::check()
- [ ] Pass devtools::test()
- [x] Tested with Exploratory

### DIFF
--- a/R/get_intercom_data.R
+++ b/R/get_intercom_data.R
@@ -1,5 +1,6 @@
 #' Get data from intercom API
-#' @param endpoint Object to access under https://api.intercom.io/
+#' @param endpoint Name of target data to access under https://api.intercom.io/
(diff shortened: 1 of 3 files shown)


R/get_intercom_data.R
```diff
@@ -1,13 +1,46 @@
 #' Get data from intercom API
-#' @param app_id App ID
-#' @param key API Key
+#' @param endpoint Object to access under https://api.intercom.io/
```
hideaki: Please add comments on possible values, or example if it is too many on this "endpoint".

double-y: Added some examples.
I meant names of objects in intercom service but updated it to " Name of target data"

PR #257, opened 2017-04-14
xgboost update

double-y:
### Description
Added na.action argument so that users can decide how to handle NA values

### Checklist

- [ ] Pass devtools::check()
- [x] Pass devtools::test()
- [x] Tested with Exploratory

### DIFF
--- a/R/build_model.R
+++ b/R/build_model.R
@@ -41,27 +41,28 @@ build_model_ <- function(data, model_func, seed = 0, test_rate = 0, group_cols =
     stop("test_rate has to be between 0 and 1")
   }
 
-  newdata <- data
+  # use this for preprocess
+  processed <- data
   # avoid name conflict of grouping columns
-  colnames(newdata)[group_col_index] <- avoid_conflict(
+  colnames(processed)[group_col_index] <- avoid_conflict(
     reserved_names,
-    colnames(newdata)[group_col_index],
+    colnames(processed)[group_col_index],
     ".group"
   )
 
   # make column names unique
-  colnames(newdata) <- make.unique(colnames(newdata), sep = "")
+  colnames(processed) <- make.unique(colnames(processed), sep = "")
 
   if(!is.null(group_cols)){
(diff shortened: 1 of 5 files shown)


R/build_model.R
```diff
@@ -41,43 +41,47 @@ build_model_ <- function(data, model_func, seed = 0, test_rate = 0, group_cols =
     stop("test_rate has to be between 0 and 1")
   }
 
+  # use this for preprocess
+  processed <- data
```
hideaki: why is this change necessary? some more explanation would help.

R/build_xgboost.R
```diff
@@ -1,12 +1,27 @@
 #' formula version of xgboost
 #' @export
```
hideaki: Explanation of na.action parameter would help here.

R/build_model.R
```diff
@@ -41,43 +41,47 @@ build_model_ <- function(data, model_func, seed = 0, test_rate = 0, group_cols =
...
+  group_col_names <- grouped_by(processed)
 
   # integrate all additional parameters
   dots <- lazyeval::all_dots(.dots, ...)
 
   # check if variables in grouped_col_names are not used
   formula <- dots$formula
   if(!is.null(formula)){
-    vars <- all.vars(formula$expr)
+    # expand formula with column names of data
```
hideaki: is this change related to na.action, or is it for something else?

R/build_model.R
double-y: Ok, I'll add a comment

R/build_model.R
double-y: Ah, this is for a different issue. dot was not handled well, so uses terms function to expand it.

R/build_xgboost.R
double-y: Ok, will add it

R/build_model.R
hideaki: Please explain that in a comment so that we accumulate the knowledge of why we do it certain way.

PR #262, opened 2017-04-24
add lists/members for backward compatibility

double-y:
### Description
* lists/members is now duplicated as "export/1.0/list" from export API but should be supported for backward compatibility

* use endpoint instead of path in access_api

### Checklist

- [ ] Pass devtools::check()
- [ ] Pass devtools::test()
- [x] Tested with Exploratory

### DIFF
--- a/R/get_mailchimp_data.R
+++ b/R/get_mailchimp_data.R
@@ -150,11 +150,11 @@ get_mailchimp_data <- function(endpoint, date_type = "exact", date_since = NULL,
       unnest_without_empty(data)
   } else if(endpoint == "lists/members"){
     # get member info from REST api for each list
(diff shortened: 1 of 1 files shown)


double-y:
Test
- [x] refresh mailchimp data from previous version of Exploratory

R/get_mailchimp_data.R
```diff
@@ -148,6 +148,20 @@ get_mailchimp_data <- function(endpoint, date_type = "exact", date_since = NULL,
     # , so unnest_without_empty is used
     ret %>%
       unnest_without_empty(data)
+  } else if(endpoint == "lists/members"){
+    # get member info from REST api for each list
```
hideaki: Please comment that this is backward compatibility for version 3.3.

double-y: Ok, added a comment

PR #266, opened 2017-04-27
Handle the case pooled connection becomes invalid (timeout etc.)

hideaki:
### Description
Handle the case pooled connection becomes invalid (timeout etc.)
1. For mongo, mysql, postgres, check validity of pooled connection before using.
2. For ODBC, where we don't know how to check validity of connection,
do pooling only when data source dialog is open. For that add switch to turn on/off pooling.

### Checklist

- [x] Pass devtools::check()
- [ ] Pass devtools::test()
- [ ] Tested with Exploratory


double-y:
looks good to me

PR #273, opened 2017-05-03
set better prophet seasonality default values

hideaki:
### Description
Set better prophet seasonality default values, to avoid the situation where forecast result degrades to only linear trend line.

### Checklist

- [ ] Pass devtools::check()
- [x] Pass devtools::test()
- [ ] Tested with Exploratory


hideaki:
- [x] tested on desktop

PR #276, opened 2017-05-11
fix pivot 

double-y:
### Description
fix https://github.com/exploratory-io/tam/issues/5731

fix https://github.com/exploratory-io/tam/issues/5739
and added empty data frame validation to other functions too

### Checklist

- [x] Pass devtools::check()
- [x] Pass devtools::test()
- [x] Tested with Exploratory


R/util.R
```diff
@@ -785,9 +790,19 @@ pivot_ <- function(df, formula, value_col = NULL, fun.aggregate = mean, fill = N
     dplyr::do_(.dots=setNames(list(~pivot_each(.)), tmp_col)) %>%
     unnest_with_drop_(tmp_col)
 
-  # replace NA values in missing columns in some groups with fill
+  if(length(rows) == 1){
```
hideaki: why is this operation limited to length(rows) == 1 case?

double-y: When rows are more than 2, it's concatenated, so the data type can't be converted

image

double-y: When rows are more than 2 columns, it's concatenated, so the type can't be converted

image

PR #280, opened 2017-05-23
add value replacement

double-y:
### Description
fix https://github.com/exploratory-io/tam/issues/5858

### Checklist

- [x] Pass devtools::check()
- [x] Pass devtools::test()
- [x] Tested with Exploratory

### DIFF
--- a/R/test_wrapper.R
+++ b/R/test_wrapper.R
@@ -40,6 +40,7 @@ do_t.test <- function(df, value, key=NULL, ...){
     ret[["alternative"]] <- as.character(ret[["alternative"]])
 
     # "two.sided" should be "two sided" because "two.sided" is confused as a url by Exploratory Desktop
+    # this can be "greater" or "lower" but they are okay.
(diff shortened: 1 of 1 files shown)


R/test_wrapper.R
```diff
@@ -39,6 +39,9 @@ do_t.test <- function(df, value, key=NULL, ...){
     ret[["method"]] <- as.character(ret[["method"]])
     ret[["alternative"]] <- as.character(ret[["alternative"]])
 
+    # "two.sided" should be "two sided" because "two.sided" is confused as a url by Exploratory Desktop
+    ret[["alternative"]] <- ifelse(ret[["alternative"]] == "two.sided", "two sided", ret[["alternative"]])
```
hideaki: What are other possible values than "two.sided" here? Can they also become a link?

double-y: They can be "greater" or "lower" too but they won't be a link

double-y: added a comment about it

PR #281, opened 2017-05-27
CausalImpact support

hideaki:
### Description
CausalImpact support

### Checklist

- [x] Pass devtools::check()
- [ ] Pass devtools::test()
- [ ] Tested with Exploratory


R/causal_impact.R
```diff
@@ -0,0 +1,186 @@
...
+#' @param niter - Number of MCMC Samples.
+#' @param standardize.data - Whether to standardize data.
+#' @param prior.level.sd - Prior Standard Deviation of Random Walk
+#' @param nseasons - Period of Seasonal Trend. e.g. 7, for weekly trend.
+#' @param season.duration - Used with nseasons. How many unit time one season consists of. e.g. 24, when unit time is hour.
+#' @param dynamic.regression - Whether to include time-varying regression coefficients.
+#' @param ... - extra values to be passed to CausalImpact::CausalImpact.
+do_causal_impact_ <- function(df, time_colname, formula, intervention_time = NULL, output_type = "series",
+                              niter = NULL, standardize.data = NULL, prior.level.sd = NULL, nseasons = NULL, season.duration = NULL, dynamic.regression = NULL, ...) {
+  y_colname <- as.character(lazyeval::f_lhs(formula))
```
double-y: Won't there be a case like log(y) ~ pred1 + pred2 ?
If it can be, as.character should be all.vars

double-y: image

R/causal_impact.R
```diff
@@ -0,0 +1,186 @@
...
+
+  if(time_colname %in% grouped_col){
+    stop(paste0(time_colname, " is grouped. Please ungroup it."))
+  }
+
+  if (!class(df[[time_colname]]) %in% c("Date", "POSIXct")) {
+    stop(paste0(time_colname, " must be Date or POSIXct."))
+  }
+
+  if (!is.null(intervention_time) && class(intervention_time) != "character" && class(df[[time_colname]]) != class(intervention_time)) {
```
double-y: I think this is easier to understand
 
!(is.null(intervention_time) || class(intervention_time) == "character" || class(df[[time_colname]]) == class(intervention_time)))

Because "!" appears only once

R/causal_impact.R
```diff
@@ -0,0 +1,186 @@
...
+  do_causal_impact_each <- function(df) {
+    # select only columns that appear in the formula.
+    # following fails with column names with spaces most likely because of dplyr bug.
+    # input_df <- dplyr::select_(df, .dots = c(time_colname, all_column_names))
+    # using base R style to avoid it.
+    # since this base R style "select" seems to mess up grouping info if done outside of do_causal_impact_each,
+    # it has to be done inside do_causal_impact_each.
+    input_df <- df[, c(time_colname, all_column_names)]
+    # rename the y column to fixed name y so that it is easier to handle in the next step.
+    input_df <- dplyr::rename_(input_df, y = y_colname)
```
double-y: Won't this y column name conflict with other column names? Like "y" with time_colname or predictor columns

R/causal_impact.R
hideaki: Will use all.vars.

R/causal_impact.R
hideaki: changed the condition that way.

R/causal_impact.R
hideaki: It's possible but I don't know the way to do  `dplyr::select(input_df, y, dplyr::everything())` without a fixed column name. Let me know if there is a better way.
For now, I renamed the name y to .y_value, which would have smaller possibility of conflict.

R/causal_impact.R
```diff
@@ -0,0 +1,186 @@
...
+
+#' NSE version of do_causal_impact_
+#' @export
+do_causal_impact <- function(df, time, formula, ...) {
+  time_colname <- col_name(substitute(time))
+  do_causal_impact_(df, time_colname, formula = formula, ...)
+}
+
+#' @param df - Data frame
+#' @param time_colname - Column that has time data
```
hideaki: OK, I thought time_colname was a better name, since it is explicit about that it is a name rather than column vector itself, but let's keep it consistent with others for now.

R/causal_impact.R
hideaki: That works. Thanks for the info!

R/causal_impact.R
```diff
@@ -0,0 +1,184 @@
...
+    time_points_vec <- df[[time_col]]
+    # select only columns that appear in the formula.
+    # following fails with column names with spaces most likely because of dplyr bug.
+    # input_df <- dplyr::select_(df, .dots = all_column_names)
+    # using base R style to avoid it.
+    # since this base R style "select" seems to mess up grouping info if done outside of do_causal_impact_each,
+    # it has to be done inside do_causal_impact_each.
+    input_df <- df[, all_column_names]
+    # bring y column at the beginning of the input_df, so that CausalImpact understand this is the column to predict.
+    move_col(input_df, y_colname, 1)
```
double-y: Shouldn't this be 
`input_df <- move_col(input_df, y_colname, 1)` 
?

hideaki: Yes! fixed. Tested it by moving date column to the last before do_causal_impact.

PR #284, opened 2017-05-30
Fill NA values in causal impact input

hideaki:
### Description
Fill NA values in causal impact input

### Checklist

- [x] Pass devtools::check()
- [ ] Pass devtools::test()
- [ ] Tested with Exploratory

### DIFF
--- a/R/causal_impact.R
+++ b/R/causal_impact.R
@@ -40,6 +40,7 @@ do_causal_impact <- function(df, time, formula, ...) {
 #'                       "StructTS" - Predict with Structural Time Series Model.
 #'                       "locf" - Fill with last previous non-NA value.
 #'                       "value" - Fill with the value of na_fill_value.
+#'                       NULL - Skip NA fill. Use this only when you know there is no NA.
 #' @param na_fill_value - Value to fill NA when na_fill_type is "value"
 #' @param output_type - Type of output data frame:
 #'                      "series" - time series (default)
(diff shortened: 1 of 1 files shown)


R/causal_impact.R
```diff
@@ -34,6 +34,13 @@ do_causal_impact <- function(df, time, formula, ...) {
 #' @param time_col - Column that has time data
 #' @param formula - Formula with target value column on the left-hand side, and predictor columns on the right-hand side. e.g. y ~ predictor1 + predictor2
 #' @param intervention_time - The point of time where intervention happened.
+#' @param na_fill_type - Type of NA fill:
```
double-y: How about making this argument na_fill and if it is numeric, regarding that it's the value to fill NA?

hideaki: Thought about that but did not do that since it would be more complex to generate the command from UI definition JSON.

double-y: ok

R/causal_impact.R
```diff
@@ -95,6 +96,22 @@ do_causal_impact_ <- function(df, time_col, formula, intervention_time = NULL, o
...
+    }
+    else if (na_fill_type == "locf") {
+      df_zoo <- zoo::na.locf(df_zoo)
+    }
+    else if (na_fill_type == "StructTS") {
+      df_zoo <- zoo::na.StructTS(df_zoo)
+    }
+    else if (na_fill_type == "value") {
+      df_zoo <- zoo::na.fill(na_fill_value)
+    }
```
double-y: No need to add else clause?
Maybe NA values should be filtered?

double-y: Or maybe it should throw an error

hideaki: Added logic to throw error.

PR #302, opened 2017-07-04
make prediction_binary on xgboost_binary work on new data without response column  (fix #301)

### LINKED ISSUE #301: new data prediction to xgboost_binary should not require a response column
### Issue Description
If new data to predict doesn't have the response column used to build xgboost_binary, this error appears

`model frame and formula mismatch in model.matrix()`

### Steps to Reproduce
Provide detail steps to reproduce the issue, including which data you use

#### Expected Behavior
What should happen

#### Actual Behavior
What really happened

##### Error Message
Provide error message or screenshot if available

double-y:
### Description
make prediction_binary on xgboost_binary work on new data without response column

Fix #301

### Checklist

- [ ] Pass devtools::check()
- [x] Pass devtools::test()
- [ ] Tested with Exploratory

### DIFF
--- a/R/broom_wrapper.R
+++ b/R/broom_wrapper.R
@@ -495,7 +495,7 @@ prediction_binary <- function(df, threshold = 0.5, ...){
   }
 
   thres <- if (!is.numeric(threshold)) {
-    # need answer to optimize threshold,
+    # need actual column to optimize threshold,
     # so the column name should be validated
     if(!actual_col %in% colnames(ret)) {
       stop("There is no actual result in data and can't optimize threshold.")
--- a/R/broom_wrapper.R
+++ b/R/broom_wrapper.R
@@ -527,7 +527,9 @@ prediction_binary <- function(df, threshold = 0.5, ...){
   } else {
     if(!is.null(df[["model"]][[1]]) && !is.null(df[["model"]][[1]]$y_levels)){
       # this is new data prediction for xgboost_binary
-      # to check levels of response column
+      # to check levels of response column because
+      # it might be factor with two levels, 2 numbers or logical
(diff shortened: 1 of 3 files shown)


R/broom_wrapper.R
```diff
@@ -520,7 +525,13 @@ prediction_binary <- function(df, threshold = 0.5, ...){
       NULL
     }
   } else {
-    NULL
+    if(!is.null(df[["model"]][[1]]) && !is.null(df[["model"]][[1]]$y_levels)){
+      # this is new data prediction for xgboost_binary
+      # to check levels of response column
```
hideaki: Was not able to understand what "check levels of response column".
Please add some more explanation.

R/build_xgboost.R
```diff
@@ -348,10 +348,18 @@ augment.xgboost_binary <- function(x, data = NULL, newdata = NULL, ...) {
       data
     }
 
+    # copy ret_data to add response_col if it doesn't exist.
+    # it's needed for model.matrix function
+    md_frame <- ret_data
```
hideaki: Looks like this md_frame is just a temporary function, but it made me thing what "md" stands for, which I'm still not sure.
Could you use more obvious name, or maybe just overwrite ret_data?

R/broom_wrapper.R
double-y: Ok, added comment

R/build_xgboost.R
double-y: ret_data should be preserved, so need to use another name

double-y: Modified it to df_for_model_matrix

PR #304, opened 2017-07-04
Make row_as_header function not to convert factor into integer

double-y:
### Description
fix https://github.com/exploratory-io/tam/issues/6095

### Checklist

- [ ] Pass devtools::check()
- [x] Pass devtools::test()
- [ ] Tested with Exploratory

### DIFF
--- a/R/toplevel_util.R
+++ b/R/toplevel_util.R
@@ -9,7 +9,9 @@ row_as_header <- function(df, row_index = 1, prefix = "", clean_names = TRUE){
 
   loadNamespace("stringr")
   loadNamespace("janitor")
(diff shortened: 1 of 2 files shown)


R/toplevel_util.R
```diff
@@ -9,7 +9,9 @@ row_as_header <- function(df, row_index = 1, prefix = "", clean_names = TRUE){
 
   loadNamespace("stringr")
   loadNamespace("janitor")
-  names <- as.character(df[row_index, ])
+  header_row <- df[row_index, ]
+  # make values to character
```
hideaki: Please add comment explaining why we do this rather than just as.character like before this change.

PR #321, opened 2017-08-02
fix group by problem feature importance 

double-y:
### Description
Fix group by problem which was caused by changing names by clean_names

### Checklist

- [ ] Pass devtools::check()
- [x] Pass devtools::test()
- [ ] Tested with Exploratory

### DIFF
--- a/R/randomForest_tidiers.R
+++ b/R/randomForest_tidiers.R
@@ -710,6 +710,10 @@ calc_feature_imp <- function(df,
 
   grouped_cols <- grouped_by(df)
 
+  if (any(c(target_col, cols) %in% grouped_cols)) {
+    stop("grouping column is used as variable columns")
+  }
+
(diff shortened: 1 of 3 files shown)


R/randomForest_tidiers.R
```diff
@@ -708,19 +708,43 @@ calc_feature_imp <- function(df,
...
+
   # randomForest fails if columns are not clean
   clean_df <- janitor::clean_names(df)
   # this mapping will be used to restore column names
   name_map <- colnames(clean_df)
   names(name_map) <- colnames(df)
 
+  # grouping column should be original column
+  # so the value of grouping columns
+  # should be still the names of grouping columns
```
hideaki: Please keep the explanation you had in the previous version that explains the behavior of clean_names. That info is needed to understand this comment.

PR #322, opened 2017-08-02
modify bayes output labels

double-y:
### Description
* modify output labels for do_bayes_ab to show blue line right hand side

image

* modify column values for this legend

image

### Checklist

- [ ] Pass devtools::check()
- [x] Pass devtools::test()
- [x] Tested with Exploratory

### DIFF
--- a/R/do_bayes_ab.R
+++ b/R/do_bayes_ab.R
@@ -237,14 +237,14 @@ tidy.bayesTest <- function(x, percentLift = 0, credInt = 0.9, type = "summary", 
     indice_b <- beta_b$x > 0 & beta_b$x < 1
 
     a_data <- data.frame(
-      ab_identifier = rep("A", length(indice_a)),
+      ab_identifier = "A",
       conversion_rate_pct = beta_a$x[indice_a] * 100,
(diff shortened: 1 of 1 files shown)


R/do_bayes_ab.R
```diff
@@ -236,26 +236,30 @@ tidy.bayesTest <- function(x, percentLift = 0, credInt = 0.9, type = "summary",
...
-      ab_identifier = ab_identifier[[1]],
+      ab_identifier = rep("A", length(indice_a)),
       conversion_rate_pct = beta_a$x[indice_a] * 100,
       probability_density = beta_a$y[indice_a],
       stringsAsFactors = FALSE
     )
 
     b_data <- data.frame(
-      ab_identifier = ab_identifier[[2]],
+      ab_identifier = rep("B", length(indice_a)),
```
hideaki: Is this indice_a supposed to be indice_b?

hideaki: BTW, why do we need rep() as opposed to just using single "B"? Please explain.

double-y: Ah, right, it automatically replicate single value, so no need for rep. Will remove it

PR #329, opened 2017-08-04
show column names if it's logical

double-y:
### Description

Avoid removing column names for TRUE FALSE

image

### Checklist

- [ ] Pass devtools::check()
- [ ] Pass devtools::test()
- [x] Tested with Exploratory


double-y:
After discussion with Hideaki, removing original column name is reverted

image

PR #332, opened 2017-08-04
add factor support

double-y:
### Description
fix https://github.com/exploratory-io/tam/issues/6427

factor support for impute_na

### Checklist

- [ ] Pass devtools::check()
- [x] Pass devtools::test()
- [ ] Tested with Exploratory

### DIFF
--- a/R/na_util.R
+++ b/R/na_util.R
@@ -47,10 +47,26 @@ impute_na <- function(target, type = mean, val = 0, ...) {
     }, value = {
       if(length(val) == 1){
         # if val is length 1, na is filled with the value
+        if(is.factor(target)) {
+          # if target is factor, val should be added as target level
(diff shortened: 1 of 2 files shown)


R/na_util.R
```diff
@@ -47,10 +47,26 @@ impute_na <- function(target, type = mean, val = 0, ...) {
     }, value = {
       if(length(val) == 1){
         # if val is length 1, na is filled with the value
+        if(is.factor(target)) {
+          # if target is factor, val should be added as target level
+          # otherwise, it doesn't replace NA
```
hideaki: What if the val is already in the levels of target?

double-y: Adding duplicated level doesn't cause error. Added it in the comment

PR #367, opened 2017-10-25
Fix: logical factor level order is reversed on confusion matrix from calc_feature_imp 

hideaki:
### Description
Set logical factor ordered as "TRUE" then "FALSE" in calc_feature_imp.

### Checklist

- [x] Pass devtools::check()
- [x] Pass devtools::test()
- [ ] Tested with Exploratory

### DIFF
--- a/R/randomForest_tidiers.R
+++ b/R/randomForest_tidiers.R
@@ -759,10 +759,17 @@ calc_feature_imp <- function(df,
   # if target is numeric, it is regression but
   # if not, it is classification
   if (!is.numeric(clean_df[[clean_target_col]])) {
-    # limit the number of levels in factor by fct_lump
-    clean_df[[clean_target_col]] <- forcats::fct_lump(
-      as.factor(clean_df[[clean_target_col]]), n = target_n
-    )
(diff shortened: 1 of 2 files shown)


R/randomForest_tidiers.R
```diff
@@ -759,10 +759,17 @@ calc_feature_imp <- function(df,
...
+    if (!is.logical(clean_df[[clean_target_col]])) {
+      # limit the number of levels in factor by fct_lump
+      clean_df[[clean_target_col]] <- forcats::fct_lump(
+        as.factor(clean_df[[clean_target_col]]), n = target_n
+      )
+    }
+    else {
+      # we need to convert logical to factor since na.roughfix only works for numeric or factor.
+      # for logical set TRUE, FALSE level order for better visualization.
+      clean_df[[clean_target_col]] <- factor(clean_df[[clean_target_col]], levels = c("TRUE", "FALSE"))
```
kei51e: You may want to handle NA too. The order should be TRUE, FALSE, NA in that case.

kei51e:
Looks good, thanks for adding test cases.

PR #388, opened 2017-12-01
better holiday dataframe and future dataframe handling

hideaki:
### Description

Support group_by on holidays data frame
Support group_by on cap data frame (future data frame)
When cap is not specified on cap data frame, use it just as future data frame to specify which data to predict.

### Checklist

- [ ] Pass devtools::check()
- [ ] Pass devtools::test()
- [ ] Tested with Exploratory


hideaki:
This includes changes for prophet 2.1 which we won't support for 4.2.

PR #405, opened 2017-12-25
smote: remember factor levels order and put it back later

hideaki:
### Description
remember factor levels order and put it back later, so that the target column has same levels and order of levels at input and output.

### Checklist

- [ ] Pass devtools::check()
- [x] Pass devtools::test()
- [x] Tested with Exploratory


hideaki:
- tested on Exploratory to check #7810 is fixed.

PR #409, opened 2017-12-29
sample before smote for performance

hideaki:
### Description
sample before smote for performance
### Checklist

- [ ] Pass devtools::check()
- [x] Pass devtools::test()
- [ ] Tested with Exploratory


kei51e:
confirmed with hideaki that we don't set seed here for now.

PR #412, opened 2018-01-02
Variable Importance: precision, recall, etc. should move to Model Quality from MQ by Class

hideaki:
### Description
Variable Importance: precision, recall, etc. should move to Model Quality from MQ by Class

### Checklist

- [ ] Pass devtools::check()
- [x] Pass devtools::test()
- [x] Tested with Exploratory


R/randomForest_tidiers.R
```diff
@@ -1044,7 +1044,8 @@ calc_feature_imp <- function(df,
 #' TODO: not really for external use. hide it.
 #' TODO: use this other places doing similar thing.
 #' @export
-evaluate_classification <- function(actual, predicted, class, pretty.name = FALSE) {
+#' @param multi_class - TRUE when we need class and data_size, which we show for multiclass classification case.
+evaluate_classification <- function(actual, predicted, class, multi_class = TRUE, pretty.name = FALSE) {
```
kei51e: Logic looks ok. But I just noticed that the variable name 'class'. It conflicts with the base function 'class' so it should use a different name I think.

hideaki: added TODO comment.

PR #415, opened 2018-01-29
Fix: Linear Regression: when variable is ordered factor, term names becomes strange

hideaki:
### Description
Fixes 3 factor related issues in build_lm.fast and build_coxph.fast below.
- Linear Regression: when variable is ordered factor, term names becomes strange
- Linear Regression: Base level of factored order variable is not respected by lm
- Analytics: We should pick the first 10 levels for Factor Columns

### Checklist

- [ ] Pass devtools::check()
- [x] Pass devtools::test()
- [ ] Tested with Exploratory

### DIFF
--- a/R/build_lm.R
+++ b/R/build_lm.R
@@ -279,7 +279,8 @@ build_lm.fast <- function(df,
           }
           df[[col]] <- NULL # drop original Date/POSIXct column to pass SMOTE later.
         } else if(is.factor(df[[col]])) {
-          # 1. if the data is ordered factor, turn it into unordered. For ordered factor,
+          # 1. if the data is factor, respect the levels and keep first 10 levels, and make others "Others" level.
+          # 2. if the data is ordered factor, turn it into unordered. For ordered factor,
           #    lm/glm takes polynomial terms (Linear, Quadratic, Cubic, and so on) and use them as variables,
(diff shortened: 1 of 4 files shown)


hideaki:
- [x] check base level of _wday variable generated from Date column on Exploratory

R/build_lm.R
```diff
@@ -257,12 +257,12 @@ build_lm.fast <- function(df,
 
           c_cols <- c(c_cols, absolute_time_col, wday_col, day_col, yday_col, month_col, year_col)
           df[[absolute_time_col]] <- as.numeric(df[[col]])
-          # turn it into character since if it is factor, the name of term is broken
-          df[[wday_col]] <- as.character(lubridate::wday(df[[col]], label=TRUE))
+          # turn it into unordered factor since if it is ordered factor, the name of term is broken
+          df[[wday_col]] <- factor(lubridate::wday(df[[col]], label=TRUE), ordered=FALSE)
```
kei51e: Why do you specify ordered=FALSE explicitly here?

hideaki: added comment.

kei51e: I'm sorry but I can't understand the reason why it needs to be unordered factor by this comment.

PR #416, opened 2018-02-02
Filter Inf/-Inf from variables of calc_feature_imp to avoid error

hideaki:
### Description
Filter Inf/-Inf from variables of calc_feature_imp to avoid error

### Checklist

- [ ] Pass devtools::check()
- [x] Pass devtools::test()
- [x] Tested with Exploratory

### DIFF
--- a/R/randomForest_tidiers.R
+++ b/R/randomForest_tidiers.R
@@ -968,7 +968,7 @@ calc_feature_imp <- function(df,
         } else {
           # filter Inf/-Inf to avoid following error from ranger.
           # Error in seq.default(min(x, na.rm = TRUE), max(x, na.rm = TRUE), length.out = length.out) : 'from' must be a finite number
-          df <- df %>% dplyr::filter(!is.infinite(df[[col]]))
+          df <- df %>% dplyr::filter(!is.infinite(.[[col]]))
         }
       }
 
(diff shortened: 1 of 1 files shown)


R/randomForest_tidiers.R
```diff
@@ -977,9 +965,28 @@ calc_feature_imp <- function(df,
           # and limit the number of levels in factor by fct_lump.
           # we need to convert logical to factor too since na.roughfix only works for numeric or factor.
           df[[col]] <- forcats::fct_explicit_na(forcats::fct_lump(as.factor(df[[col]]), n=predictor_n, ties.method="first"))
+        } else {
+          # filter Inf/-Inf to avoid following error from ranger.
+          # Error in seq.default(min(x, na.rm = TRUE), max(x, na.rm = TRUE), length.out = length.out) : 'from' must be a finite number
+          df <- df %>% dplyr::filter(!is.infinite(df[[col]]))
```
kei51e: Feeling a bit weird because `df[[col]]` part actually ignores the data frame passed over pipe. 

Either
```
df <- df %>% dplyr::filter(!is.infinite(.[[col]]))
```
or 
```
df <- filter(df, !is.infinite(df[[col]]))
```
would be more reasonable to me.

PR #417, opened 2018-02-09
log rank test/chi-squared test

hideaki:
### Description
Add support for log rank test/chi-squared test

### Checklist

- [ ] Pass devtools::check()
- [x] Pass devtools::test()
- [x] Tested with Exploratory


R/survival.R
```diff
@@ -0,0 +1,125 @@
...
+  grouped_col <- grouped_by(df)
+
+  if (!is.null(substitute(cohort))) {
+    cohort_col <- col_name(substitute(cohort))
+    if (!is.null(cohort_func)) {
+      df[[cohort_col]] <- extract_from_date(df[[cohort_col]], type=cohort_func)
+    }
+  }
+  else {
+    cohort_col <- "1"
```
kei51e: What is this "1" for?

R/survival.R
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  # substitute is needed because time can be
+  # NSE column name and it throws an evaluation error
```
kei51e: What is NSE?

hideaki: It stands for Non-Standard Evaluation. Added comment.

R/survival.R
hideaki: Added comment on it.

PR #418, opened 2018-02-09
Call `index()` via namespace (zoo::)

IronistM:
### Description
Using the function `do_anomaly_detection()` fails because `zoo` package is not loaded. Since namespace loading is prefereed this is now prepended to the function.

### Checklist

- [x] Pass devtools::check()
- [x] Pass devtools::test()
- [x] Tested with Exploratory

Extra test cases are not strictly needed, IMO.


hideaki:
Thank you for the PR. Merged it!

PR #426, opened 2018-03-01
Make t-test/ANOVA more robust in case of group_by

hideaki:
### Description
In exp_ttext and exp_anova, we will keep going even if one of the group_by group has error, so that we can show the result from other groups.

### Checklist

- [ ] Pass devtools::check()
- [x] Pass devtools::test()
- [x] Tested with Exploratory


kei51e:
Looks good. Thanks for adding tests.

PR #429, opened 2018-03-12
Skip normality test for a column with always same value

hideaki:
### Description
- Skip normality test for a column with always same value
- Turn off Yate's correction for exp_chisq by default

### Checklist

- [ ] Pass devtools::check()
- [x] Pass devtools::test()
- [x] Tested with Exploratory

### DIFF
--- a/R/test_wrapper.R
+++ b/R/test_wrapper.R
@@ -548,7 +548,7 @@ exp_normality<- function(df, ..., n_sample = 50) {
     df.model <- data.frame()
 　  for (col in selected_cols) {
       if (n_distinct(df[[col]], na.rm=TRUE) <= 1) {
-        # skip if the column has only 1 unique value, to avoid error.
+        # skip if the column has only 1 unique value or only NAs, to avoid error.
         # TODO: show what happened in the summary table.
       }
       else {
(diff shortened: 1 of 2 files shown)


R/test_wrapper.R
```diff
@@ -542,34 +547,40 @@ exp_normality<- function(df, ..., n_sample = 50) {
...
-　    df.qqline <- dplyr::bind_rows(df.qqline, data.frame(x=c(min_x, max_x), refline_y=c(ref_min_y,ref_max_y), col=col))
-
-      if (n_sample > 5000) {
-        n_sample <- 5000 # shapiro.test takes only up to max of 5000 samples. 
-      }
-
-      if (length(df[[col]]) > n_sample) {
-        col_to_test <- sample(df[[col]], n_sample)
-        sample_size <- n_sample
+      if (n_distinct(df[[col]], na.rm=TRUE) <= 1) {
```
kei51e: Is there a case n_distinct(x) <1  ?

hideaki: It can happen when all values are NAs.

hideaki: Added comment.

kei51e: Would be great if you can add some comment on that.

PR #440, opened 2018-05-20
Add and export %in_or_all% operator

hideaki:
### Description
Add and export %in_or_all% operator

### Checklist

- [ ] Pass devtools::check()
- [ ] Pass devtools::test()
- [x] Tested with Exploratory


hideaki:
When vector with length more than 1 and a vector with length 1 is mixed in calculation, R treats the one with length 1 as if it has the same length as the other vector.
So I think leaving it length 1 should work most of the cases, and is more efficient.

kei51e:
We should follow the spec of what match function (%in%). 

image

I think we can just reverse the result of NULL case for this case.

```
> 1:10 %in% c(1,6)
 [1]  TRUE FALSE FALSE FALSE FALSE  TRUE FALSE FALSE FALSE FALSE
> 1:10 %in% NULL
 [1] FALSE FALSE FALSE FALSE FALSE FALSE FALSE FALSE FALSE FALSE
> !(1:10 %in% NULL)
 [1] TRUE TRUE TRUE TRUE TRUE TRUE TRUE TRUE TRUE TRUE
```

hideaki:
OK, I changed it so that it returns same length vector.

PR #441, opened 2018-05-22
chisq: preserve logical TRUE/FALSE order on chart

hideaki:
### Description
To preserve logical TRUE/FALSE order on chart, output should be casted back to logical when input columns were logical originally.

### Checklist

- [ ] Pass devtools::check()
- [x] Pass devtools::test()
- [x] Tested with Exploratory


kei51e:
Looks good. thanks for adding tests.

PR #444, opened 2018-05-24
do_cmdscale : handle all 0 distances case

hideaki:
### Description
Handle the case where all distances are 0.

### Checklist

- [ ] Pass devtools::check()
- [x] Pass devtools::test()
- [x] Tested with Exploratory

### DIFF
--- a/R/stats_wrapper.R
+++ b/R/stats_wrapper.R
@@ -228,6 +228,12 @@ do_cmdscale_ <- function(df,
 
   # this is executed on each group
   do_cmdscale_each <- function(df){
+    if (all(df[[value_col]]==0)) {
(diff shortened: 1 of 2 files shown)


R/stats_wrapper.R
```diff
@@ -228,6 +228,12 @@ do_cmdscale_ <- function(df,
 
   # this is executed on each group
   do_cmdscale_each <- function(df){
+    if (all(df[[value_col]]==0)) {
```
kei51e: What about other corner cases such as all NA, all Inf/-Inf case? Are those already taken care of? or it shouldn't happen at there?

kei51e: Thanks for checking out.

PR #453, opened 2018-06-04
[MongoDB] fix wrong default value for timeout.

hidekoji:
### Description

fix wrong default value for getMongoCollectionNumberOfRows

### Checklist

- [ ] Pass devtools::check()
- [ ] Pass devtools::test()
- [ ] Tested with Exploratory

### DIFF
--- a/R/system.R
+++ b/R/system.R
@@ -300,7 +300,7 @@ getMongoCollectionNames <- function(host = "", port = "", database = "", usernam
 getMongoCollectionNumberOfRows <- function(host = NULL, port = "", database = "",
                                            username = "", password = "", collection = "",
                                            isSSL=FALSE, authSource=NULL, cluster = NULL, additionalParams = "",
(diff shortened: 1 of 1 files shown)


R/system.R
```diff
@@ -300,7 +300,7 @@ getMongoCollectionNames <- function(host = "", port = "", database = "", usernam
 getMongoCollectionNumberOfRows <- function(host = NULL, port = "", database = "",
                                            username = "", password = "", collection = "",
                                            isSSL=FALSE, authSource=NULL, cluster = NULL, additionalParams = "",
-                                           timeout = timeout, ...){
+                                           timeout = "", ...){
```
hideaki: Why does it have to be an empty string as opposed to NULL, which seems like more natural choice provided this is a numeric argument?

hidekoji: Updated as suggested.

PR #458, opened 2018-06-11
do_prophet: handle "month" and larger time unit in test mode

hideaki:
### Description
do_prophet: handle "month" and larger time unit in test mode correctly.

### Checklist

- [ ] Pass devtools::check()
- [x] Pass devtools::test()
- [x] Tested with Exploratory

### DIFF
--- a/R/prophet.R
+++ b/R/prophet.R
@@ -153,8 +153,11 @@ do_prophet_ <- function(df, time_col, value_col = NULL, periods, time_unit = "da
 
 
     if (test_mode) {
-      # fill aggregated_data$ds with missing data/time.
-      # this is necessary to make forecast period correspond with test period in test mode when there is missing date/time in original aggregated_data$ds.
+      # Remove end of aggregated_data as test data to make training data.
+
+      # Fill aggregated_data$ds with missing data/time.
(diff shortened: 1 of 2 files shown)


R/prophet.R
```diff
@@ -165,9 +165,18 @@ do_prophet_ <- function(df, time_col, value_col = NULL, periods, time_unit = "da
       else {
         time_unit_for_seq <- time_unit
       }
-      ts <- seq.POSIXt(as.POSIXct(min(aggregated_data$ds)), as.POSIXct(max(aggregated_data$ds)), by=time_unit_for_seq)
-      if (lubridate::is.Date(aggregated_data$ds)) {
-        ts <- as.Date(ts)
+      # Create periodical sequence of time to fill missing date/time
+      if (time_unit %in% c("hour", "minute", "second")) { # Use seq.POSIXt for unit smaller than day.
```
kei51e: Do you need the same processing in non-test mode?

hideaki: This is only for determining when to start test period. So no need of this for non-test mode.

hideaki: Added some more comment to explain what this is for.

PR #466, opened 2018-07-25
Use openxlsx package for importing .xlsx file

hidekoji:
### Description

For better memory usage point of view, use openxlsx package for importing/exporting .xlsx file.

### Checklist

- [ ] Pass devtools::check()
- [ ] Pass devtools::test()
- [ ] Tested with Exploratory

### DIFF
--- a/R/system.R
+++ b/R/system.R
@@ -1446,9 +1446,9 @@ read_excel_file <- function(path, sheet = 1, col_names = TRUE, col_types = NULL,
   loadNamespace('stringr')
   df <- NULL
   # for .xlsx file extension
-  if(stringr::str_detect(path, '.xlsx')) {
+  if(stringr::str_detect(path, '\\.xlsx')) {
     if(n_max != Inf) {
-      df <- openxlsx::read.xlsx(xlsxFile = path, rows=skip+1:n_max, sheet = sheet, colNames = col_names, na.strings = na, skipEmptyRows = skipEmptyRows, skipEmptyCols = skipEmptyCols , check.names = check.names, detectDates = detectDates)
+      df <- openxlsx::read.xlsx(xlsxFile = path, rows=(skip+1):n_max, sheet = sheet, colNames = col_names, na.strings = na, skipEmptyRows = skipEmptyRows, skipEmptyCols = skipEmptyCols , check.names = check.names, detectDates = detectDates)
     } else {
       df <- openxlsx::read.xlsx(xlsxFile = path, sheet = sheet, colNames = col_names, startRow = skip+1, na.strings = na, skipEmptyRows = skipEmptyRows, skipEmptyCols = skipEmptyCols, check.names = check.names, detectDates = detectDates)
     }
--- a/R/system.R
+++ b/R/system.R
@@ -1477,11 +1477,20 @@ read_excel_file <- function(path, sheet = 1, col_names = TRUE, col_types = NULL,
   df
 }
 
-#'Wrapper for openxlsx::getSheetNames to support remote file
+#'Wrapper for readxl::excel_sheets to support remote file
 #'@export
 get_excel_sheets <- function(path){
-  loadNamespace("openxlsx")
-  openxlsx::getSheetNames(path)
+  loadNamespace("readxl")	+  loadNamespace("openxlsx")
+  loadNamespace("stringr")	+  openxlsx::getSheetNames(path)
+  if (stringr::str_detect(path, "^https://") ||
+      stringr::str_detect(path, "^http://") ||
+      stringr::str_detect(path, "^ftp://")) {
+    tmp <- download_data_file(path, "excel")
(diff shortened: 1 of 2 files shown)


hideaki:
- Does this work with remote file?
- Please add some comment on why we use this package as opposed to readxl.

R/system.R
```diff
@@ -1437,20 +1437,44 @@ download_data_file <- function(url, type){
...
-    readxl::read_excel(path, sheet = sheet, col_names = col_names, col_types = col_types, na = na, trim_ws = trim_ws, skip = skip, n_max = n_max)
+read_excel_file <- function(path, sheet = 1, col_names = TRUE, col_types = NULL, na = "", skip = 0, trim_ws = TRUE, n_max = Inf, detectDates = FALSE, skipEmptyRows = FALSE, skipEmptyCols = FALSE, check.names = FALSE, ...){
+  loadNamespace("openxlsx")
+  loadNamespace('readr')
+  loadNamespace('stringr')
+  df <- NULL
+  # for .xlsx file extension
+  if(stringr::str_detect(path, '.xlsx')) {
+    if(n_max != Inf) {
+      df <- openxlsx::read.xlsx(xlsxFile = path, rows=c(skip+1:n_max), sheet = sheet, colNames = col_names, na.strings = na, skipEmptyRows = skipEmptyRows, skipEmptyCols = skipEmptyCols , check.names = check.names, detectDates = detectDates)
```
hideaki: Is the c() around skip+1:n_max necessary?
Also, I'm not very sure of priority of operators here. Maybe I can be more confident with (skip+1):n_max.

hidekoji: removed `c()` wrap.

hidekoji:
It supports remote file (URL).  I added some comment for the reason why we want to use openxlsx::read.xlsx.

R/system.R
```diff
@@ -1437,20 +1437,45 @@ download_data_file <- function(url, type){
...
-  } else {
-    # if it's local file simply call readxl::read_excel
-    readxl::read_excel(path, sheet = sheet, col_names = col_names, col_types = col_types, na = na, trim_ws = trim_ws, skip = skip, n_max = n_max)
+read_excel_file <- function(path, sheet = 1, col_names = TRUE, col_types = NULL, na = "", skip = 0, trim_ws = TRUE, n_max = Inf, detectDates = FALSE, skipEmptyRows = FALSE, skipEmptyCols = FALSE, check.names = FALSE, ...){
+  loadNamespace("openxlsx")
+  loadNamespace('readr')
+  loadNamespace('stringr')
+  df <- NULL
+  # for .xlsx file extension
+  if(stringr::str_detect(path, '.xlsx')) {
```
hideaki: This regex would match when it should not match.
image

R/system.R
```diff
@@ -1437,20 +1437,45 @@ download_data_file <- function(url, type){
...
+    if(n_max != Inf) {
+      df <- openxlsx::read.xlsx(xlsxFile = path, rows=skip+1:n_max, sheet = sheet, colNames = col_names, na.strings = na, skipEmptyRows = skipEmptyRows, skipEmptyCols = skipEmptyCols , check.names = check.names, detectDates = detectDates)
```
hideaki: I think you need parenthesis around skip+1 as I doubted in my previous comment.
image

R/system.R
hidekoji: changed as suggested

R/system.R
hidekoji: changed as suggested

R/system.R
```diff
@@ -1440,13 +1440,13 @@ download_data_file <- function(url, type){
 #'Wrapper for openxlsx::read.xlsx (in case of .xlsx file) and readxl::read_excel (in case of old .xls file)
 #'Use openxlsx::read.xlsx since it's memory footprint is less than that of readxl::read_excel and this creates benefit for users with less memory like Windows 32 bit users.
 #'@export
-read_excel_file <- function(path, sheet = 1, col_names = TRUE, col_types = NULL, na = "", skip = 0, trim_ws = TRUE, n_max = Inf, detectDates = FALSE, skipEmptyRows = FALSE, skipEmptyCols = FALSE, check.names = FALSE, ...){
+read_excel_file <- function(path, sheet = 1, col_names = TRUE, col_types = NULL, na = "", skip = 0, trim_ws = TRUE, n_max = Inf, useOld = FALSE, detectDates = FALSE, skipEmptyRows = FALSE, skipEmptyCols = FALSE, check.names = FALSE, ...){
```
hideaki: "Old" is a relative term we should avoid using since what is "old" could change in future.
Something like use_readxl would be a better name.
Also, snake cases and camel cases are mixed in arguments of this function.
For parameters we are adding from now, let's use snake cases.

hidekoji: Changed argument name to use_readxl

PR #472, opened 2018-08-28
Use glue instead of getOptlong

hideaki:
### Description
Use glue instead of getOptlong for parameter binding in SQL and mongo js query.

### Checklist

- [ ] Pass devtools::check()
- [ ] Pass devtools::test()
- [x] Tested with Exploratory

### DIFF
--- a/R/system.R
+++ b/R/system.R
@@ -249,6 +249,21 @@ odbc_glue_transformer <- function(code, envir) {
   glue::collapse(val, sep=", ")
 }
 
+bigquery_glue_transformer <- function(code, envir) {
+  val <- eval(parse(text = code), envir)
+  if (is.character(val) || is.factor(val)) {
+    # escape for Standard SQL for bigquery
+    # https://cloud.google.com/bigquery/docs/reference/standard-sql/lexical
+    # escape for js
+    val <- gsub("\\", "\\\\", val, fixed=TRUE)
+    val <- gsub("\"", "\\\"", val, fixed=TRUE)
+    val <- paste0('"', val, '"')
+  }
+  # TODO: How should we handle logical, Date, POSIXct, POSIXlt?
+  #       Does expression like 1e+10 work?
+  glue::collapse(val, sep=", ")
+}
+
 #' @export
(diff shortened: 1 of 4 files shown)


R/system.R
```diff
@@ -215,14 +215,61 @@ getMongoURL <- function(host = NULL, port, database, username, pass, isSSL=FALSE
   return (url)
 }
 
+# glue transformer for mongo js query.
+# supports character, factor, logical, Date, POSIXct, POSIXlt, and numeric.
+# TODO: do we need to handle NA?
```
kei51e: How do you handle NA?

kei51e: can you add testcase for those cases?

R/system.R
```diff
@@ -215,14 +215,64 @@ getMongoURL <- function(host = NULL, port, database, username, pass, isSSL=FALSE
...
+
+odbc_glue_transformer <- function(code, envir) {
+  val <- eval(parse(text = code), envir)
+  if (is.character(val) || is.factor(val)) {
+    # escape for SQL
+    # TODO: check if this makes sense for Dremio and Athena
+    val <- gsub("'", "''", val, fixed=TRUE) # both Oracle and SQL Server escapes single quote by doubling them.
+    val <- paste0("'", val, "'") # both Oracle and SQL Server quotes strings with single quote.
+  }
+  # TODO: How should we handle logical, Date, POSIXct, POSIXlt?
```
kei51e: I think we need NA handling here too.

R/system.R
```diff
@@ -215,14 +215,64 @@ getMongoURL <- function(host = NULL, port, database, username, pass, isSSL=FALSE
...
+  #       Does expression like 1e+10 work?
+  glue::collapse(val, sep=", ")
+}
+
+bigquery_glue_transformer <- function(code, envir) {
+  val <- eval(parse(text = code), envir)
```
kei51e: I think we need NA handling here too.

R/system.R
hideaki: Added TODO comment.

R/system.R
hideaki: Added TODO comment.

PR #475, opened 2018-09-01
prophet regressor and holiday support

hideaki:
### Description
prophet regressor and holiday support

### Checklist

- [ ] Pass devtools::check()
- [ ] Pass devtools::test()
- [x] Tested with Exploratory


R/prophet.R
```diff
@@ -2,10 +2,11 @@
 
 #' NSE version of do_prophet_
 #' @export
-do_prophet <- function(df, time, value = NULL, ...){
+do_prophet <- function(df, time, value = NULL, periods, holiday = NULL, ...){
```
kei51e: There's no default value for periods and no na check on periods. I think you should have at least either one.

PR #478, opened 2018-09-09
Fix for cox regression prediction with probability at specific time

hideaki:
### Description
Fix for cox regression prediction with probability at specific time.
Fix for #9878

### Checklist

- [ ] Pass devtools::check()
- [ ] Pass devtools::test()
- [x] Tested with Exploratory


R/broom_wrapper.R
```diff
@@ -578,7 +578,7 @@ prediction_coxph <- function(df, time = NULL, threshold = 0.5, ...){
   # get group columns.
   # we assume that columns of model df other than the ones with reserved name are all group columns.
   model_df_colnames = colnames(df)
-  group_by_names <- model_df_colnames[!model_df_colnames %in% c("source.data", ".test_index", "model")]
+  group_by_names <- model_df_colnames[!model_df_colnames %in% c("source.data", ".test_index", "model", ".model_metadata")]
```
kei51e: If we already know the column names to groupby here, we should go with the white list approach instead of the black list approach in order to avoid the same problem in future.

kei51e: If not, please refer where to look for where those columns are defined in a comment.

hideaki: Added comment.

PR #495, opened 2018-10-05
ttest: filter NA if it is the 3rd category

hideaki:
### Description
Filter NA if it is the 3rd category, to make it 2-category data.

### Checklist

- [ ] Pass devtools::check()
- [ ] Pass devtools::test()
- [x] Tested with Exploratory

### DIFF
--- a/R/test_wrapper.R
+++ b/R/test_wrapper.R
@@ -353,8 +353,9 @@ exp_ttest <- function(df, var1, var2, func2 = NULL, ...) {
     df <- df %>% dplyr::mutate(!!rlang::sym(var2_col) := extract_from_date(!!rlang::sym(var2_col), type=func2))
   }
   
-  if (n_distinct(df[[var2_col]]) != 2) {
-    if (n_distinct(df[[var2_col]], na.rm=TRUE) == 2) { # automatically filter NA to make number of category 2, if it is the 3rd category.
+  n_distinct_res <- n_distinct(df[[var2_col]]) # save n_distinct result to avoid repeating the relatively expensive call.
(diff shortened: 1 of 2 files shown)


R/test_wrapper.R
```diff
@@ -350,11 +350,16 @@ exp_ttest <- function(df, var1, var2, func2 = NULL, ...) {
   grouped_cols <- grouped_by(df)
 
   if (!is.null(func2) && (is.Date(df[[var2_col]]) || is.POSIXct(df[[var2_col]]))) {
-    df <- df %>% mutate(!!rlang::sym(var2_col) := extract_from_date(!!rlang::sym(var2_col), type=func2))
+    df <- df %>% dplyr::mutate(!!rlang::sym(var2_col) := extract_from_date(!!rlang::sym(var2_col), type=func2))
   }
   
   if (n_distinct(df[[var2_col]]) != 2) {
-    stop(paste0("Variable Column (", var2_col, ") has to have 2 kinds of values."))
+    if (n_distinct(df[[var2_col]], na.rm=TRUE) == 2) { # automatically filter NA to make number of category 2, if it is the 3rd category.
```
kei51e: I remember that distinct/n_distinct is a relatively heavy operation. You can check NA in line 357  instead. Here's the benchmark. 8-10 times faster.

image

hideaki: But we want to do this only when removing NA makes it 2 categories.
If there is only NA and something else, we don't want to do that.

kei51e: You can keep the result of the first n_distinct. You can check only n=3 case. Am I missing somthing?

PR #499, opened 2018-10-24
Make sure to skip file(file) handle for exploratory::read_delim_file when file path is all ascii characters

hidekoji:
### Description

Make sure to skip file(file) handle for exploratory::read_delim_file when file path is all ascii characters

### Checklist

- [ ] Pass devtools::check()
- [ ] Pass devtools::test()
- [ ] Tested with Exploratory

### DIFF
--- a/R/system.R
+++ b/R/system.R
@@ -1573,9 +1573,9 @@ read_delim_file <- function(file, delim, quote = '"',
     # if it's local file simply call readr::read_delim
     # reading through file() is to be able to read files with path that includes multibyte chars.
     # without it, error is thrown from inside read_delim.
-    file_path = file
+    file_path <- file
     if(stringi::stri_enc_mark(file) != "ASCII"){
-      file_path = file(file)
+      file_path <- file(file)
     }
     readr::read_delim(file_path, delim, quote = quote, escape_backslash = escape_backslash, escape_double = escape_double, col_names = col_names, col_types = col_types,
                       locale = locale, na = na, quoted_na = quoted_na, comment = comment, trim_ws = trim_ws, skip = skip, n_max = n_max, guess_max = guess_max, progress = progress)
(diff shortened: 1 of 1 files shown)


R/system.R
```diff
@@ -1573,8 +1573,13 @@ read_delim_file <- function(file, delim, quote = '"',
     # if it's local file simply call readr::read_delim
     # reading through file() is to be able to read files with path that includes multibyte chars.
     # without it, error is thrown from inside read_delim.
-    readr::read_delim(file(file), delim, quote = quote, escape_backslash = escape_backslash, escape_double = escape_double, col_names = col_names, col_types = col_types,
-                      locale = locale, na = na, quoted_na = quoted_na, comment = comment, trim_ws = trim_ws, skip = skip, n_max = n_max, guess_max = guess_max, progress = progress)
+    if(stringi::stri_enc_mark(file) == "ASCII"){
+      readr::read_delim(file, delim, quote = quote, escape_backslash = escape_backslash, escape_double = escape_double, col_names = col_names, col_types = col_types,
```
hideaki: The only difference here is the first argument. I don't think you need to repeat the whole function call here.

R/system.R
```diff
@@ -1573,8 +1573,13 @@ read_delim_file <- function(file, delim, quote = '"',
     # if it's local file simply call readr::read_delim
     # reading through file() is to be able to read files with path that includes multibyte chars.
     # without it, error is thrown from inside read_delim.
-    readr::read_delim(file(file), delim, quote = quote, escape_backslash = escape_backslash, escape_double = escape_double, col_names = col_names, col_types = col_types,
+    file_path = file
```
hideaki: We use <- rather than =.

R/system.R
```diff
@@ -1573,8 +1573,13 @@ read_delim_file <- function(file, delim, quote = '"',
...
+    if(stringi::stri_enc_mark(file) != "ASCII"){
+      file_path = file(file)
```
hideaki: We use <- rather than =.

R/system.R
hidekoji: changed as suggested.

R/system.R
hidekoji: changed as suggested.

PR #501, opened 2018-11-03
To avoid the issue that bind_rows fails due to column data type mismatch, first load each csv file as character then do type_convert after bind_rows

hidekoji:
### Description

When downloading CSV files from cloud storage, it's possible that `readr:read_csv` guesses data type for the same column in a different way per file.  To work around this, first treat all the column as a character, then perform bind_rows and finally do type_convert for column type guessing.

### Checklist

- [ ] Pass devtools::check()
- [ ] Pass devtools::test()
- [ ] Tested with Exploratory

### DIFF
--- a/R/system.R
+++ b/R/system.R
@@ -928,7 +928,7 @@ downloadDataFromGoogleCloudStorage <- function(bucket, folder, download_dir, tok
   });
   files <- list.files(path=download_dir, pattern = ".gz");
   # pass progress as FALSE to prevent SIGPIPE error on Exploratory Desktop.
-  # To avoid the issue that bind_rows thorws an error due to column data type mismatch,
+  # To avoid the issue that bind_rows throws an error due to column data type mismatch,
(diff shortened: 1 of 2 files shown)


R/system.R
```diff
@@ -928,7 +928,7 @@ downloadDataFromGoogleCloudStorage <- function(bucket, folder, download_dir, tok
   });
   files <- list.files(path=download_dir, pattern = ".gz");
   # pass progress as FALSE to prevent SIGPIPE error on Exploratory Desktop.
-  df <- lapply(files, function(file){readr::read_csv(stringr::str_c(download_dir, "/", file), progress = FALSE)}) %>% dplyr::bind_rows()
+  df <- lapply(files, function(file){readr::read_csv(stringr::str_c(download_dir, "/", file), col_types = readr::cols(.default = "c"), progress = FALSE)}) %>% dplyr::bind_rows() %>% readr::type_convert()
```
hideaki: Please add comment on why we do this.

PR #502, opened 2018-12-11
Amazon AWS connection support

hidekoji:
### Description

Added two new APIs:

- getAWSAthenaConnection
- clearAWSAthenaConnection
- queryAWSAthena

### Checklist

- [ ] Pass devtools::check()
- [ ] Pass devtools::test()
- [ ] Tested with Exploratory


R/system.R
```diff
@@ -368,6 +368,58 @@ getMongoCollectionNumberOfRows <- function(host = NULL, port = "", database = ""
   return(result)
 }
 
+#' Returns a AWS Athena connection.
+#' @export
+getAWSAthenaConnection <- function(driver = "", region = "", authenticationType = "IAM Credentials", s3OutputLocation = "", user = "", password = "", additionalParams = "", ...) {
+  loadNamespace("RODBC")
+  loadNamespace("DBI")
```
hideaki: Do we need DBI when using RODBC?

hidekoji: DBI is actually not used so removed this loadNamespace for DBI.

PR #508, opened 2018-12-28
NA fill options after aggregation for do_prophet

hideaki:
### Description
Support NA fill options done after aggregation for do_prophet.

### Checklist

- [ ] Pass devtools::check()
- [ ] Pass devtools::test()
- [x] Tested with Exploratory


R/anomaly_detection.R
```diff
@@ -164,7 +164,7 @@ do_anomaly_detection_ <- function(
         time = time_col_values
       ) %>%
         dplyr::group_by(time) %>%
-        dplyr::summarise(count = n())
+        dplyr::summarise(val = n())
```
kei51e: What's the reason of changing the name from count to val?

PR #511, opened 2018-12-30
fix Google BigQuery : Error: Invalid value at 'max_results.value' (TYPE_UINT32), "1e+06" [badRequest]

hidekoji:
### Description

Currently, `getGoogleBigQueryTables` passes `page_size=1000000` when it calls 

```r
    tables <- bigrquery::list_tables(project, dataset, page_size=1000000);
```

However, this started raising below error:

```r
Error: Invalid value at 'max_results.value' (TYPE_UINT32), "1e+06" [badRequest]
```
So it needs to decrease default value to `page_size=10000`

### Checklist

- [ ] Pass devtools::check()
- [ ] Pass devtools::test()
- [x] Tested with Exploratory

### DIFF
--- a/R/system.R
+++ b/R/system.R
@@ -1208,7 +1208,7 @@ getGoogleBigQueryTables <- function(project, dataset, tokenFileId=""){
     token <- getGoogleTokenForBigQuery(tokenFileId);
     bigrquery::set_access_cred(token)
     # if we do not pass max_results, it only returnss 50 items. so explicitly set it.
-    tables <- bigrquery::list_tables(project, dataset, page_size=1000000);
+    tables <- bigrquery::list_tables(project, dataset, page_size=10000);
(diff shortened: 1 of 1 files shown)


R/system.R
```diff
@@ -1208,7 +1208,7 @@ getGoogleBigQueryTables <- function(project, dataset, tokenFileId=""){
     token <- getGoogleTokenForBigQuery(tokenFileId);
     bigrquery::set_access_cred(token)
     # if we do not pass max_results, it only returnss 50 items. so explicitly set it.
```
hideaki: Does max_results here mean page_size?
Also, please add comment on what you found about the error you are fixing in this PR.

hidekoji: I added a comment about the error I got from Google BigQuery API call and my observation.

PR #514, opened 2019-01-06
exp_kmeans sample_n error

hideaki:
### Description
- Fix exp_kmeans sample_n error by checking n is actually smaller than original number of rows.
- Update Feature Importance's max_pd_vars default from 10 to 12. (Looks better on 3 or 4 column facet.)

### Checklist

- [ ] Pass devtools::check()
- [ ] Pass devtools::test()
- [x] Tested with Exploratory - Verified that the issue does not happen with the reproducible case.


hideaki:
add exploratory:: namespace to usages of sample_rows().

PR #520, opened 2019-01-27
Complete survival curve data

hideaki:
### Description
Complete survival curve data so that survival table does not have missing cells.

### Checklist

- [ ] Pass devtools::check()
- [ ] Pass devtools::test()
- [x] Tested with Exploratory

### DIFF
--- a/R/survival.R
+++ b/R/survival.R
@@ -129,7 +129,7 @@ tidy.survfit_exploratory <- function(x, ...) {
   ret <- broom:::tidy.survfit(x, ...)
 
   # for better viz, add time=0 row for each group when it is not already there.
-  add_time_zero_row_each <- function(df) {
+  complete_times_each <- function(df) {
     if (nrow(df[df$time==0,]) == 0) { # do this only when time=0 row is not already there.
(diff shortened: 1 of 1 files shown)


R/survival.R
```diff
@@ -128,17 +128,21 @@ exp_survival <- function(df, time, status, start_time = NULL, end_time = NULL, e
...
-  # for better viz, add time=0 row for each group when it is not already there.
-  add_time_zero_row_each <- function(df) {
+  # for line chart and pivot table, add time=0 row when it is not already there, and rows for other missing times for each group.
+  complete_times_each <- function(df) {
     if (nrow(df[df$time==0,]) == 0) { # do this only when time=0 row is not already there.
       df <- rbind(data.frame(time=0, n.risk=df$n.risk[1], n.event=0, n.censor=0, estimate=1, std.error=0, conf.high=1, conf.low=1), df)
     }
+    # Fill data for missing period, as if there was a row for 0 event and 0 censor.
+    df <- df %>% tidyr::complete(time=0:max(time))
+    df <- df %>% fill_between(n.risk, estimate, std.error, conf.high, conf.low)
```
kei51e: Can this fill the empty values at the very top or bottom of the data frame?

hideaki: That's a good point. I just thought about it and think we'd better keep it this way. Added comment.

PR #522, opened 2019-02-04
Wrapper for Analytics View for Boruta

hideaki:
### Description
Add wrapper for Analytics View for Boruta, exp_boruta().

### Checklist

- [ ] Pass devtools::check()
- [x] Pass devtools::test()
- [x] Tested with Exploratory


tests/testthat/test_boruta.R
```diff
@@ -0,0 +1,19 @@
+# how to run this test:
+# devtools::test(filter="rpart")
+
+if (!exists("flight")) {
+  # To skip repeated data loading, run the following outside of the context of the test,
+  # so that it stays even after the test.
+  flight <- exploratory::read_delim_file("https://www.dropbox.com/s/f47baw5f3v0xoll/airline_2013_10_tricky_v3.csv?dl=1", ",", quote = "\"", skip = 0 , col_names = TRUE , na = c("","NA") , locale=readr::locale(encoding = "UTF-8", decimal_mark = "."), trim_ws = FALSE , progress = FALSE) %>% exploratory::clean_data_frame()
```
kei51e: Does this test data include multibyte characters/languages?

R/randomForest_tidiers.R
```diff
@@ -1749,3 +1749,133 @@ tidy.rpart <- function(x, type = "importance", pretty.name = FALSE, ...) {
...
+
+  do_on_each_group(clean_df, each_func, name = "model", with_unnest = FALSE)
+}
+
+tidy.Boruta <- function(x, ...) {
+  res <- tidyr::gather(as.data.frame(x$ImpHistory), "variable","importance")
+  decisions <- data.frame(variable=names(x$finalDecision), decision=x$finalDecision)
+  res <- res %>% dplyr::left_join(decisions, by = "variable") 
+  res$variable <- x$terms_mapping[res$variable] # Map variable names back to original.
+  res <- res %>% dplyr::filter(decision %in% c("Confirmed", "Tentative", "Rejected"))
```
kei51e: Should describe why we need to filter by those 3 items. Also, seems it is used in multiple places so it would be better to define a constant and reuse it.

hideaki: Added comment. The list of levels appear in line 1869 and 1871, but in 1871, they has to be directly embedded in arguments. So let me keep it this way.

tests/testthat/test_boruta.R
hideaki: We take care of it in a separate test suite.

PR #533, opened 2019-02-19
SMOTE related improvements

hideaki:
### Description
- Use new set of arguments for exp_balance(), so that we don't down-sample precious minority data.
- Handle ordered factor columns
- Add seed argument on exp_balance()
- Call new exp_balance() from Analytics View functions.
- Add "synthesized" column to show synthesized rows.

### Checklist

- [ ] Pass devtools::check()
- [ ] Pass devtools::test()
- [ ] Tested with Exploratory


R/randomForest_tidiers.R
```diff
@@ -684,16 +684,188 @@ rf_partial_dependence <- function(df, ...) { # TODO: write test for this.
...
+    ret
+  }
+
+  if (is.null(target_size)) {
+    if (minority_size / (minority_size + majority_size) >= target_minority_perc / 100) {
+      # Already enough minority for the ratio even without SMOTE.
+      # No Action. Already above target minority ratio.
+      newdataset <- data %>% dplyr::mutate(synthesized = FALSE)
+    }
+    else if (minority_size * (100 + max_synth_perc) / 100 / (minority_size * (100 + max_synth_perc) / 100 + majority_size) >= target_minority_perc / 100) {
```
hidekoji: this if /else condition is little tricky since it does a lot of calculation in it.  Some comment explaining what's actually is calculated in the condition help to understand it. Same for all the other similar if/else conditions.

PR #546, opened 2019-03-09
make sure to pass "n" argument to get a full result for schema list

hidekoji:
### Description

We need to pass `n` argument like `DBI::dbFetch(resultSet, n = n)` otherwise, we only get first 500 rows.

### Checklist

- [x] Pass devtools::check()
- [x] Pass devtools::test()
- [x] Tested with Exploratory

### DIFF
--- a/R/system.R
+++ b/R/system.R
@@ -793,14 +793,14 @@ getListOfColumns <- function(type, host, port, databaseName, username, password,
 
 #' API to execute a query that can be handled with DBI
 #' @export
-executeGenericQuery <- function(type, host, port, databaseName, username, password, query, catalog = "", schema = "", n = -1){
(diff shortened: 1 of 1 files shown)


R/system.R
```diff
@@ -793,14 +793,14 @@ getListOfColumns <- function(type, host, port, databaseName, username, password,
 
 #' API to execute a query that can be handled with DBI
 #' @export
-executeGenericQuery <- function(type, host, port, databaseName, username, password, query, catalog = "", schema = ""){
+executeGenericQuery <- function(type, host, port, databaseName, username, password, query, catalog = "", schema = "", n = -1){
```
hideaki: Other similar function in this file uses "numOfRows" as the name for this parameter. Let't make it consistent.

hidekoji: Changed argument name to `numOfRows`

PR #547, opened 2019-03-19
Calculate average marginal effect based on pre-SMOTE data

hideaki:
### Description
Calculate average marginal effect based on pre-SMOTE data, even if model is based on SMOTEd data.

### Checklist

- [ ] Pass devtools::check()
- [ ] Pass devtools::test()
- [x] Tested with Exploratory


kei51e:
According to Hideaki, it is covered by the auto test against the function that calls this new function.

PR #548, opened 2019-03-20
support case insensitive join

hidekoji:
### Added wrapper join functions for followings:

- dplyr::left_join
- dplyr::right_join
- dplyr::full_join
- dplyr::inner_join
- dplyr::semi_join
- dplyr::anti_join

And these wrapper function supports `ignorecase` arugment.

### Checklist

- [x] Pass devtools::check()
- [x] Pass devtools::test()
- [x] Tested with Exploratory


R/join.R
```diff
@@ -0,0 +1,89 @@
+#' @ref{https://gist.github.com/jimhester/a060323a05b40c6ada34}
+#' @export
+insensitive_join <- function(fun = dplyr::left_join, type = "LEFT") {
```
hideaki: by or other args are not defined.

R/join.R
```diff
@@ -0,0 +1,89 @@
...
+  if(ignorecase) {
+    insensitive_join(dplyr::right_join, "RIGHT")(x, y, by = by, copy = copy, suffix = suffix)
+  } else {
+    dplyr::right_join(x, y, by = by, copy = copy, suffix = suffix);
+  }
+}
+
+#' @export
+#' Wrapper function for dplyr's full_join to support case insensitive join.
+full_join <- function(x, y, by = NULL, copy = FALSE, suffix = c(".x", ".y"), ignorecase = FALSE, ......) {
```
hideaki: Why is this with 6 dots?

R/join.R
hidekoji: this is helper function which changes the behavior of the join function so argument by or other arguments are not required. as you can see below, by and other arguments are handled inside `substitute` like below.

```
body(new_fun) <- substitute({
    by <- dplyr:::common_by(by, x, y)
``

R/join.R
hidekoji: removed unnecessary `...`

PR #554, opened 2019-04-07
WIP: Use new APIs of the text2vec (0.5.1)

htamakos:
### Description
- Replace `text2vec::get_idf` to `text2vec::TfIdf` class
- Add deprecated warnings for `idf_log_scale` which is an argument of `do_tfidf` function.
  - `text2vec::TfIdf` has a function for calculation of TF-IDF, named `fit_transform` or `transform`, which internally calculates IDF by private function `get_idf`. This has **NO** parameter of log_scale. 
  - So there is no way to set log scale function to `text2vec::TfIdf`. 
  - For compatibility, idf_log_scale argument is available, but log function is used no matter what function is used for the argument. Also, in that case, a warning to that effect is output.

REF: [https://rdrr.io/cran/text2vec/man/TfIdf.html](https://rdrr.io/cran/text2vec/man/TfIdf.html)
### Checklist

- [ ] Pass devtools::check()
- [x] Pass devtools::test()
- [ ] Tested with Exploratory

### DIFF
--- a/R/string_operation.R
+++ b/R/string_operation.R
@@ -190,7 +190,7 @@ do_tokenize <- function(df, input, token = "words", keep_cols = FALSE,  drop = T
 }
 
 #' Get idf for terms
-calc_idf <- function(group, term, log_scale = log, smooth_idf = FALSE){
+calc_idf <- function(group, term, smooth_idf = FALSE){
   loadNamespace("Matrix")
   loadNamespace("text2vec")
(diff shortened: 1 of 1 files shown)


kei51e:

hideaki:
Will merge this for 5.2 or 5.1.6. For now, I'm adding WIP prefix.

htamakos:
duplicated #571

PR #564, opened 2019-05-01
Work around for ranger error against parse_number output

hideaki:
# Description
- Work around for ranger error against parse_number output.
- Avoid deprecated way of calling UQ() with namespace like rlang::UQ().

# Checklist

- [ ] Pass devtools::check()
- [x] Pass devtools::test()
- [x] Tested with Exploratory

### DIFF
--- a/R/string_operation.R
+++ b/R/string_operation.R
@@ -399,7 +399,10 @@ parse_number <- function(text, ...){
   if(is.numeric(text)) {
     text
   } else {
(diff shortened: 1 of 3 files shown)


htamakos:
@hideaki 
I think that your fix itself is not a problem, but since the CI build has failed, please fix it.
I do not know why failing because I have no authority to see what Jenkins is doing.

PR #571, opened 2019-05-11
Use new APIs of the text2vec (0.5.1)

htamakos:
### Description
- Replace `text2vec::get_idf` to `text2vec::TfIdf` class
- Add deprecated warnings for `idf_log_scale` which is an argument of `do_tfidf` function.
  - `text2vec::TfIdf` has a function for calculation of TF-IDF, named `fit_transform` or `transform`, which internally calculates IDF by private function `get_idf`. This has **NO** parameter of log_scale. 
  - So there is no way to set log scale function to `text2vec::TfIdf`. 
  - For compatibility, idf_log_scale argument is available, but log function is used no matter what function is used for the argument. Also, in that case, a warning to that effect is output.

REF: [https://rdrr.io/cran/text2vec/man/TfIdf.html](https://rdrr.io/cran/text2vec/man/TfIdf.html)
### Checklist

- [ ] Pass devtools::check()
- [x] Pass devtools::test()
- [ ] Tested with Exploratory

### DIFF
--- a/R/string_operation.R
+++ b/R/string_operation.R
@@ -262,7 +262,7 @@ calc_tf_ <- function(df, group_col, term_col, weight="ratio"){
 #' @param group Column of group names
 #' @param term Column of terms
 #' @param [DEPRECATED] idf_log_scale
-#' Function to scale IDF. “log” function is always applied to IDF. Setting other functions is ignored
+#' Function to scale IDF. 'log' function is always applied to IDF. Setting other functions is ignored
 #' @export
 do_tfidf <- function(df, group, term, idf_log_scale = log, tf_weight="raw", norm="l2"){
   validate_empty_data(df)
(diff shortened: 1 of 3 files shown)


R/string_operation.R
```diff
@@ -258,9 +261,8 @@ calc_tf_ <- function(df, group_col, term_col, weight="ratio"){
 #' @param df Data frame which has columns of groups and their terms
 #' @param group Column of group names
 #' @param term Column of terms
-#' @param idf_log_scale
-#' Function to scale IDF. It might be worth trying log2 or log10.
-#' log10 strongly suppress the increase of idf values and log2 does it more weakly.
+#' @param [DEPRECATED] idf_log_scale
+#' Function to scale IDF. “log” function is always applied to IDF. Setting other functions is ignored
```
hideaki: Please use straight quote instead of curly quotes.

R/string_operation.R
```diff
@@ -272,6 +274,10 @@ do_tfidf <- function(df, group, term, idf_log_scale = log, tf_weight="raw", norm
     stop("norm argument must be l1, l2 or FALSE")
   }
 
+  if(!missing(idf_log_scale)){
+	warnings("Argument idf_log_scale is deprecated. Log is always applied to IDF.")
```
hideaki: Please use 2 spaces as indentation.

kei51e:

PR #576, opened 2019-05-30
Add str_extract_inside function to extract text inside character like bracket.

hidekoji:
# Description

Add `exploratory::str_extract_inside` that extract text inside bracket.(or other characters)

# Checklist

- [x] Pass devtools::check()
- [x] Pass devtools::test()
- [x] Tested with Exploratory


R/string_operation.R
```diff
@@ -425,3 +425,11 @@ parse_logical <- function(text, ...){
   }
 }
 
+#'Function to extract text inside the characters like bracket.
+#'@export
+str_extract_inside <- function(column, begin = "(", end = ")") {
+  exp = stringr::str_c("(?<=\\", begin, ").*?(?=\\", end, ")")
```
hideaki: Some reference url about lookahead match and lookbehind match like this as comment would help. https://stackoverflow.com/questions/3926451/how-to-match-but-not-capture-part-of-a-regex

I think the ".*?" in the center of regex can be just ".*".
"?" looks redundant if I'm not missing anything.

hideaki: It seems that the ? after asterisk has effect to handle this case. We'd better keep it. Comment about it would help too.

```
> stringr::str_extract("aaa[xxx][ggg]","(?<=\\[).*(?=\\])")
[1] "xxx][ggg"
> stringr::str_extract("aaa[xxx][ggg]","(?<=\\[).*?(?=\\])")
[1] "xxx"
```

hidekoji: Added comment.

hideaki: Please add comment about shortest match too. This is about the ? I was talking in the above comment.
https://stackoverflow.com/questions/1919982/regex-smallest-possible-match-or-nongreedy-match

hidekoji: Added comment.

hideaki:
Looks only first character of begin/end is escaped in regex, and this won't work with the case like begin="[[", and = "]]". Also, not sure something like begin="begin", end="end" would work.

hidekoji:
Added validation for begin and end argument and also added test cases for these.

PR #590, opened 2019-06-18
fix Data Source: Could not open an SPSS file and got No method asJSON S3 class: haven_labelled error

hidekoji:
# Description

fix Data Source: Could not open an SPSS file and got No method asJSON S3 class: haven_labelled error

Since jsonlite cannot handle haven_labelled class, need to convert it as a factor.

# Checklist

- [x] Pass devtools::check()
- [x] Pass devtools::test()
- [x] Tested with Exploratory

### DIFF
--- a/R/system.R
+++ b/R/system.R
@@ -1317,7 +1317,8 @@ scrape_html_table <- function(url, index, heading, encoding = NULL) {
 #' see https://github.com/exploratory-io/tam/issues/1481
 #' @export
 handleLabelledColumns = function(df){
-  is_labelled <- which(lapply(df, class) == "labelled")
+  # check if column class is labelled, have_labelled, etc
+  is_labelled <- which(stringr::str_detect(lapply(df, class), "labelled"))
   df[is_labelled] <- lapply(df[is_labelled], haven::as_factor)
   df
 }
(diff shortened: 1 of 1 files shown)


R/system.R
```diff
@@ -1317,7 +1317,8 @@ scrape_html_table <- function(url, index, heading, encoding = NULL) {
 #' see https://github.com/exploratory-io/tam/issues/1481
```
hideaki: Found typo above this line - factoror

R/system.R
```diff
@@ -1317,7 +1317,8 @@ scrape_html_table <- function(url, index, heading, encoding = NULL) {
...
 #' @export
 handleLabelledColumns = function(df){
-  is_labelled <- which(lapply(df, class) == "labelled")
+  # check if column class is labelled, haven_labelled, etc
+  is_labelled <- which(stringr::str_detect(lapply(df, class), "labelled"))
```
hideaki: How about just matching with "labelled" and "haven_labelled" if that are all the possibilities we know, to avoid false positive?
Also, some more comment on the context where this is called would help.

R/system.R
hidekoji: fixed typo.

R/system.R
hidekoji: changed to check both "labelled" and "haven_labelled"

hidekoji: Also added the comment that explains why we need this function.

PR #591, opened 2019-06-18
Support summarize_group function which is wrapper function to dplyr's group by and summarize

hidekoji:
# Description

Added the summarize_group function which is a wrapper function that handles dplyr's group by and summarize as a single step.

# Checklist

- [x] Pass devtools::check()
- [x] Pass devtools::test()
- [x] Tested with Exploratory


tests/testthat/test_util.R
```diff
@@ -836,3 +836,11 @@ test_that("get_mode", {
   res <- get_mode(x, na.rm = TRUE)
   expect_equal(as.character(res), "2019-01-02")
 })
+
+test_that("summarize_group", {
+ df <- mtcars %>% exploratory::summarize_group(grp_cols = c("cyl", "mpg"), grp_aggregations = c("none", "mean"), count = n())
```
hideaki: grp_aggregations argument is not really about aggregation. It's rather categorization.
Also, abbreviating "group" to "grp" only in the argument names while the function name is summarize_group looks a little inconsistent to me. How about "group_cols" and "group_funs" instead of "grp_cols" and "grp_aggregations"?

R/util.R
```diff
@@ -1262,31 +1262,73 @@ do_on_each_group_2 <- function(df, func1, func2, params1 = quote(list()), params
...
+}
+
 #' @export
 extract_from_date <- function(x, type = "fltoyear") {
   ret <- NULL
   switch(type,
     fltoyear = {
       ret <- lubridate::floor_date(x, unit="year")
     },
+    rtoyear = {
```
hideaki: Looks same as fltoyear. If you need this, please comment on reason.

R/util.R
```diff
@@ -1711,3 +1780,82 @@ get_mode <- function(x, na.rm = FALSE) {
   ux <- unique(x)
   return(ux[which.max(tabulate(match(x, ux)))])
 }
+# Wrapper function that takes care of dplyr::group_by and dplyr::summarize as a single step.
```
hideaki: Comment on arguments would help.

R/util.R
```diff
@@ -1711,3 +1780,82 @@ get_mode <- function(x, na.rm = FALSE) {
...
+# @export
+summarize_group <- function(.data, grp_cols = NULL, grp_aggregations = NULL, ...){
+  library(dplyr)
+  if(length(grp_cols) == 0) {
+    .data %>% summarize(...)
+  } else {
+    groupby_args <- list() # default empty list
+    name_list <- list()
+    name_index = 1
+    # If group by columns and associated aggregation funcionts are provided,
```
hideaki: Typo: funcionts
Those functions are rather for categorization, and not for aggregation.
"group by columns" => "group_by columns"

R/util.R
```diff
@@ -1711,3 +1780,82 @@ get_mode <- function(x, na.rm = FALSE) {
...
+            "wdaylong",
+            "hour",
+            "minute",
+            "second")) {
+          # For date column, call extract_from_date
+          rlang::quo(extract_from_date(UQ(rlang::sym(cname)), type = UQ(func)))
+        } else if (func %in% c("asnum","asint","asintby10","aschar")) {
+          # For numeric column, call categorize_numeric
+          rlang::quo(categorize_numeric(UQ(rlang::sym(cname)), type = UQ(func)))
+        } else {
```
hideaki: Comment what is this else clause is for.

R/util.R
```diff
@@ -1711,3 +1780,82 @@ get_mode <- function(x, na.rm = FALSE) {
...
+        if(is.na(func) || length(func)==0 || func == "none"){
+          cname
+        } else {
+          stringr::str_c(cname, func, sep = "_")
+        }
+      })
+      names(groupby_args) <- name_list
+      .data %>% dplyr::group_by(!!!groupby_args) %>% summarize(...)
+    } else {
+      if(!is.null(grp_cols)) {
```
hideaki: Short comment on what this case does would help.

R/util.R
hidekoji: Added comment.

tests/testthat/test_util.R
hidekoji: Changed argument name as suggested.

R/util.R
hidekoji: Added the reason for the key.

R/util.R
hidekoji: Added comments on arguments.

R/util.R
hidekoji: Added comment.

R/util.R
hidekoji: Fixed typo and updated comment.

R/util.R
hidekoji: Added comment.

R/util.R
```diff
@@ -1262,31 +1262,73 @@ do_on_each_group_2 <- function(df, func1, func2, params1 = quote(list()), params
...
+}
+
 #' @export
 extract_from_date <- function(x, type = "fltoyear") {
   ret <- NULL
   switch(type,
     fltoyear = {
       ret <- lubridate::floor_date(x, unit="year")
     },
+    rtoyear = {# This key is required by Exploratory Desktop for Summarize Group Dialog.
```
hideaki: The point I wanted comment here was,
1. This is same as fltoyear, but it is intentional.
2. The reason why Summary Group Dialog use this different name.

hidekoji: Updated comment.

PR #593, opened 2019-06-25
Add str_logical function for logical value parsing wrapper function

hidekoji:
# Description

Add str_logical function.

if you pass `true_value` argument,  if the column has the value (case insensitive) it returns TRUE.
and other FALSE. If the column value is NA it's treated as NA.

```r
str_logical(colname,  true_value = "Sign Up")
```

If you don't pass this argument, it automatically detects TRUE if:

- The column contains "yes" or "no" (case insensitive), "yes' is treated as TRUE. (NA value remains NA)
- The column contains "true" or "false" (case insensitive), "true' is treated as TRUE. (NA value remains NA)
- The column contains "1" or "0", "1' is treated as TRUE. (NA value remains NA)
- The column contains 1 or 0, 1 is treated as TRUE. (NA value remains NA)

# Checklist

- [x] Pass devtools::check()
- [x] Pass devtools::test()
- [x] Tested with Exploratory


hidekoji:
Tested performance on Cancelled Column of Tricky Flight Data (535,344 Rows), and it takes

R< try({`airline_2013_10_tricky_v3_Source1.temp` <- `airline_2013_10_tricky_v3_Source1` %>% mutate(calculation_1 = "true", `CAN CELLED` = str_logical(`CAN CELLED`));if(!is.data.frame(`airline_2013_10_tricky_v3_Source1.temp`)){stop("Error : Result must be a data frame")}; jsonlite::toJSON(TRUE); })

211 ms elapsed

PR #594, opened 2019-06-25
Pivot: fixed issue passing  date column along with lubridate function fails.

hidekoji:
# Description

This PR addresses the issue that passing Date column along with lubridate function fails

```r
df %>% pivot(lubridate::year(`FL DATE`) ~ `CAR RIER`)
```

When you pass Date (or POSIXct) column along with lubridate function that returns non-Date (or non-POSIXct) value like above, it fails with below error.

```r
Error in charToDate(x) : 
  character string is not in a standard unambiguous format
```

# Checklist

- [x] Pass devtools::check()
- [x] Pass devtools::test()
- [x] Tested with Exploratory


R/util.R
```diff
@@ -335,11 +335,25 @@ same_type <- function(vector, original){
   } else if(is.integer(original)){
     as.integer(vector)
   } else if(inherits(original, "Date")) {
-    # when original data is Date
-    as.Date(vector, tz = lubridate::tz(original))
+    # when original data is Date.
+    # if the column is wrapped with lubridate funciton, it's possible that data is converted to numeric like year
```
hideaki: Typo: funciton

hideaki: Not too sure why in the first place this function has to be called in such case. Isn't just simply not calling it an option?

R/util.R
```diff
@@ -1185,6 +1199,20 @@ na_ratio <- function(x){
   sum(is.na(x)) / length(x)
 }
 
+#' Ratio of TRUE in a vector
+#' @param x vector
+#' @export
+true_ratio <- function(x){
+  sum(x, na.rm =!all(is.na(x))) / length(x)
```
hideaki: Please add comment on intention of na.rm =!all(is.na(x)) somewhere in the code.

PR #597, opened 2019-06-27
Added new type "daytype" to extract_from_date

hidekoji:
# Description

extract_from_date now has new type "daytype" that returns Weekend if the day is Sat or Sun and Weekday anything else.

# Checklist

- [x] Pass devtools::check()
- [x] Pass devtools::test()
- [x] Tested with Exploratory

### DIFF
--- a/R/util.R
+++ b/R/util.R
@@ -1437,7 +1437,7 @@ extract_from_date <- function(x, type = "fltoyear") {
     wdaylong = {
       ret <- lubridate::wday(x, label=TRUE, abbr=FALSE)
     },
-    daytype = {
+    wdaytype = {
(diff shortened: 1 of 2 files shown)


R/util.R
```diff
@@ -1437,6 +1437,11 @@ extract_from_date <- function(x, type = "fltoyear") {
     wdaylong = {
       ret <- lubridate::wday(x, label=TRUE, abbr=FALSE)
     },
+    daytype = {
```
hideaki: The name "daytype" sounds too generic for this concept. How about "wdaytype"?

hidekoji: updated as wdaytype

PR #598, opened 2019-06-27
fix GoogleBigQuery API for query UTF8 handling issue.

hidekoji:
# Description

This PR addresses the issue that query string passed to exploratory::executeGoogleBigQuery is not properly converted to UTF 

# Checklist

- [ ] Pass devtools::check()
- [x] Pass devtools::test()
- [x] Tested with Exploratory


R/util.R
```diff
@@ -1439,11 +1439,6 @@ extract_from_date <- function(x, type = "fltoyear") {
     wdaylong = {
       ret <- lubridate::wday(x, label=TRUE, abbr=FALSE)
     },
-    wdaytype = {
```
hideaki: Don't you need to keep wdaytype option here to make chart etc. keep working?

hidekoji: Chart/Analytics/Summarize dialog use `weekend` instead of `extact_from_date` function so it works.

hidekoji: Bring back weekend.

PR #602, opened 2019-07-02
Updated str_logical to use as.logical for numeric columns

hidekoji:
# Description

For numeric columns changed to use `as.logical` 

# Checklist

- [x] Pass devtools::check()
- [x] Pass devtools::test()
- [x] Tested with Exploratory


R/string_operation.R
```diff
@@ -462,6 +462,8 @@ str_logical <- function(column, true_value = NULL) {
    # if true_value is explicitly provided, honor it
    if(!is.null(true_value)) {
      stringr::str_to_lower(stringr::str_trim(column)) == stringr::str_to_lower(true_value)
+   } else if (is.numeric(column)) { # for numeric columns (integer, double, numeric, etc), use as.logical
+     as.logical(column)
    } else { # default handling.
       # if value is either "true", "yes", "1", or 1, return TRUE
```
hideaki: I don’t think you need numeric 1 or 0 here since numeric is already handled.
(Also, it you put 1 in character vector, it is casted to character.)

R/string_operation.R
```diff
@@ -462,12 +462,14 @@ str_logical <- function(column, true_value = NULL) {
...
```
hideaki: This comment should not mention numerical  1 either.

hidekoji: Thank you for checking. Updated comment.

PR #607, opened 2019-07-06
summarize group: Supported the case where previous step has group columns

hidekoji:
# Description

Supported the case where the previous step has group columns.

# Checklist

- [x] Pass devtools::check()
- [x] Pass devtools::test()
- [x] Tested with Exploratory

### DIFF
--- a/R/util.R
+++ b/R/util.R
@@ -1932,7 +1932,7 @@ summarize_group <- function(.data, group_cols = NULL, group_funs = NULL, ...){
   if(length(group_cols) == 0) {
     .data %>% summarize(...)
   } else {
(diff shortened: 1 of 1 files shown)


R/util.R
```diff
@@ -1932,6 +1932,9 @@ summarize_group <- function(.data, group_cols = NULL, group_funs = NULL, ...){
   if(length(group_cols) == 0) {
     .data %>% summarize(...)
   } else {
+    # if group_cols argument is passed, make sure to ungroup first so that it won't thrown an error
```
hideaki:
```suggestion
    # if group_cols argument is passed, make sure to ungroup first so that it won't throw an error
```

PR #614, opened 2019-07-10
Update is_alphabet to handle non-ascii characters properly.

hidekoji:
# Description

Switched regular expression so that is_alphabet returns FALSE for non-ASCII characters.

# Checklist

- [x] Pass devtools::check()
- [x] Pass devtools::test()
- [x] Tested with Exploratory

### DIFF
--- a/R/string_operation.R
+++ b/R/string_operation.R
@@ -46,7 +46,7 @@ is_alphabet <- function(word){
   loadNamespace("stringr")
   # To treat non-ascii characters as FALSE, use [a-zA-Z]
   # instead of [:alpha:]
-  stringr::str_detect(word, "^[[a-zA-Z]]+$")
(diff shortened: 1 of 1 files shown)


R/string_operation.R
```diff
@@ -44,7 +44,9 @@ is_digit <- function(word){
 #' @export
 is_alphabet <- function(word){
   loadNamespace("stringr")
-  stringr::str_detect(word, "^[[:alpha:]]+$")
+  # To treat non-ascii characters as FALSE, use [a-zA-Z]
+  # instead of [:alpha:]
+  stringr::str_detect(word, "^[[a-zA-Z]]+$")
```
hideaki: Shouldn’t it be "^[a-zA-Z]+$"

hidekoji: Thank you for checking it. Fixed the syntax.

PR #620, opened 2019-07-31
Use correct ranger training prediction

hideaki:
# Description
- As training prediction result from ranger, use prediction explicitly done with training data, instead of the one extracted from model, which is a worse prediction than the one done explicitly.
- Avoid edarf error from too small a sample size
- Correct wrong labels of metrics from glance.ranger and glance.randomForest
- Correct handling of which category becomes TRUE for ranger/rpart

# Checklist

- [ ] Pass devtools::check()
- [ ] Pass devtools::test()
- [x] Tested with Exploratory

### DIFF
--- a/R/broom_wrapper.R
+++ b/R/broom_wrapper.R
@@ -650,7 +650,18 @@ prediction_binary <- function(df, threshold = 0.5, ...){
   }
 
   actual_val <- ret[[actual_col]]
-  actual_logical <- as.logical(as.numeric(actual_val))
+  actual_logical <- if ((is.character(actual_val) || is.factor(actual_val)) && "ranger" %in% class(first_model)) {
+    # For binary prediction of ranger/rpart used from Analytics View, we use prediction() rather than this prediction_binary().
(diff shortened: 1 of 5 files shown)


R/broom_wrapper.R
```diff
@@ -694,8 +705,13 @@ prediction_binary <- function(df, threshold = 0.5, ...){
       factor(levels(actual_val)[as.numeric(predicted) + 1], levels(actual_val))
     }
   } else if (is.character(actual_val)) {
-    # modify actual_val to factor with levels used in training data
-    if(!is.null(first_model$model) && !is.null(first_model$model[[actual_col]])){
+    if ("ranger" %in% class(first_model)) {
+      if_else(predicted, levels(first_model$y)[[1]], levels(first_model$y)[[2]])
+      lev <- levels(first_model$y)
+      factor(lev[2 - as.numeric(predicted)], lev)
```
hideaki: Add comment.

hideaki: Added comment.

PR #638, opened 2019-08-18
Pivot: fix the issue that NA case is displayed as 0

hidekoji:
# Description

fix Pivot Calculation for NA value

Before the fix, it shows 0 as a result of aggregation where it should be NA.

image

After the fix.

image

# Checklist

- [x] Pass devtools::check()
- [ ] Pass devtools::test()
- [x] Tested with Exploratory


hideaki:
I don't think this is an issue. For Number of Rows, 0 makes better default than NA.

kanaugust:
It should be NA because the 'Fill Value' is set to NA.

hideaki:
Argument default should be NA, or fill=NA should be specified explicitly at call from UI.

hidekoji:
changed argument default for fill as NA

PR #644, opened 2019-08-23
Record sampling for PCA

hideaki:
# Description
- Record sampling for PCA
- Remove unnecessary PCA from kmeans in elbow method mode.

# Checklist

- [ ] Pass devtools::check()
- [ ] Pass devtools::test()
- [x] Tested with Exploratory

### DIFF
--- a/R/kmeans.R
+++ b/R/kmeans.R
@@ -82,7 +82,7 @@ exp_kmeans <- function(df, ...,
                           trace = trace,
                           normalize_data = normalize_data,
                           seed=NULL) # Seed is already done in do_prcomp. Skip it.
-    ret <- data.frame(model = I(list(ret))) # Follow current output format for now. TODO: Revisit and support group_by.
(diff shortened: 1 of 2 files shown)


R/kmeans.R
```diff
@@ -63,29 +64,25 @@ exp_kmeans <- function(df, ...,
...
+    kmeans_df <- df %>% dplyr::select(!!!rlang::quos(...))
+    ret <- iterate_kmeans(kmeans_df,
+                          max_centers = max_centers,
+                          iter.max = iter.max,
+                          nstart = nstart,
+                          algorithm = algorithm,
+                          trace = trace,
+                          normalize_data = normalize_data,
+                          seed=NULL) # Seed is already done in do_prcomp. Skip it.
+    ret <- data.frame(model = I(list(ret))) # Follow current output format for now. TODO: Revisit and support group_by.
```
kei51e: why we need I here?

hideaki:
```suggestion
    ret <- data.frame(model = I(list(ret))) # Follow current output format for now. I() is to avoid unwanted expansion of list at creation of data frame. TODO: Revisit and support group_by.
```

hideaki: Added comment.

PR #651, opened 2019-09-19
Support for VIF, normalization, and outlier filtering

hideaki:
# Description
Support for VIF, normalization, and outlier filtering

# Checklist

- [ ] Pass devtools::check()
- [ ] Pass devtools::test()
- [x] Tested with Exploratory


R/build_lm.R
```diff
@@ -633,6 +677,22 @@ build_lm.fast <- function(df,
         }
       }
 
+      tryCatch({
+        model$vif <- car::vif(model)
+      }, error = function(e){
+        # in case of perfect multicollinearity, vif throws error with message "there are aliased coefficients in the model".
+        # Check it it is the case. If coef() includes NA, corresponding variable is causing perfect multicollinearity.
```
kei51e:
```suggestion
        # Check if it is the case. If coef() includes NA, corresponding variable is causing perfect multicollinearity.
```

kei51e: What about Inf/-Inf case?

R/build_lm.R
```diff
@@ -871,6 +931,17 @@ xlevels_to_base_level_table <- function(xlevels) {
...
+    ret <- x$vif %>% as.data.frame() %>%  tibble::rownames_to_column(var="term") %>% rename(VIF=GVIF)
+  }
+  else {
+    ret <- data.frame(term=names(x$vif), VIF=x$vif)
+  }
+  ret
+}
+
 #' special version of tidy.lm function to use with build_lm.fast.
 #' @export
```
kei51e: Should describe that it returns an empty data frame in case of error.

R/build_lm.R
```diff
@@ -1006,6 +1086,15 @@ tidy.glm_exploratory <- function(x, type = "coefficients", pretty.name = FALSE,
         dplyr::ungroup()
       ret
     },
+    vif = {
+      if (!is.null(x$vif) && "error" %nin% class(x$vif)) {
+        ret <- vif_to_dataframe(x)
+      }
+      else {
+        ret <- data.frame() # Skip output for this group. TODO: Report error in some way.
```
kei51e: Should describe that it returns an empty data frame in case of error at the beginning of the function.

R/build_lm.R
hideaki: Added comment.

R/build_lm.R
hideaki: Added comment.

PR #676, opened 2019-11-23
Add week wrapper function

kei51e:
# Description

Add `week` wrapper function. 

```
# week of year
exploratory::week(dates)
# week of quarter
exploratory::week(dates, unit="quarter")
# week of month
exploratory::week(dates, unit="month")
```

# Checklist

- [ ] Pass devtools::check()
- [x] Pass devtools::test()
- [x] Tested with Exploratory


R/util.R
```diff
@@ -2051,6 +2051,27 @@ sample_frac <- function(..., seed = NULL){
 }
 
 # Get the week number of month https://stackoverflow.com/a/58370031
+# @deprecated
```
kei51e:
```suggestion
# @deprecated Leave it for a while for Desktop 5.4.0.12 support.
```

hideaki:
lubridate::week(date) counts number of complete  7 day periods regardless of day of week, but it seems exploratory::week with unit of month and quarter is written to return week that starts from Sunday. I think we should  be consistent with lubridate::week(date) at least by default, and return number of complete  7 day periods since the beginning of month or quarter.

kei51e:
Updated the calculation logic.

PR #709, opened 2020-02-06
lm/glm: Map column name to names like c1, c2... before building models to avoid column name related issues

hideaki:
# Description
Map column name to names like c1, c2... before building models to avoid column name related issues.

# Checklist

- [ ] Pass devtools::check()
- [ ] Pass devtools::test()
- [x] Tested with Exploratory


kei51e:
Can you check the unit test diffs?

kei51e:
Hideaki will look at the test diffs later.

PR #742, opened 2020-03-02
Support MS SQL Server Data Source without DSN using odbc package

hidekoji:
# Description

Support MS SQL Server Data Source without DSN using odbc package

# Checklist

- [x] Pass devtools::check()
- [ ] Pass devtools::test()
- [x] Tested with Exploratory

### DIFF
--- a/R/system.R
+++ b/R/system.R
@@ -652,6 +652,10 @@ getDBConnection <- function(type, host = NULL, port = "", databaseName = "", use
     loadNamespace("RODBC")
     connect <- function() {
       if(type == "mssqlserver") {
+        # if platform is Linux use predefined one
+        if(Sys.info()["sysname"]=="Linux"){
+          driver <-  "ODBC Driver 17 for SQL Server";
+        }
         connstr <- stringr::str_c("Driver={", driver, "}", sep="")
         if(host != "") {
           connstr <- stringr::str_c(connstr, ";Server=tcp:", host, sep="")
--- a/R/system.R
+++ b/R/system.R
@@ -712,8 +712,10 @@ getDBConnection <- function(type, host = NULL, port = "", databaseName = "", use
     if(Sys.info()["sysname"]=="Linux"){
       driver <-  "ODBC Driver 17 for SQL Server";
     }
-    # mssqlserver uses odbc package instead of RODBC to allow users to schedule the data source
-    # on server side, so make sure odbc package is installed.
+    # mssqlserver uses odbc package instead of RODBC for two reasons:
+    # 1) The ODBC Driver 17 for SQL Server driver does not work on Mac with RODBC.
+    # 2) To allow users to schedule the data source on server side.
+    # So make sure odbc package is installed.
     if(!requireNamespace("DBI")){stop("package DBI must be installed.")}
     if(!requireNamespace("odbc")){stop("package odbc must be installed.")}
     key <- paste("mssqlserver", host, port, databaseName, username, sep = ":")
@@ -1085,7 +1087,10 @@ queryODBC <- function(dsn="", username, password, additionalParams="", numOfRows
     clearDBConnection(type, NULL, NULL, NULL, username, dsn = dsn, additionalParams = additionalParams)
     stop(err)
   })
-  # mssqlserver uses result set so make sure clear the result set.
+  # mssqlserver uses odbc package and DBI package
+  # and it gets result set with DBI package.
+  # So make sure clear the result set.
+  # For RDOBC based case, it does not use result set.
   if(type == "mssqlserver") {
     DBI::dbClearResult(resultSet)
   }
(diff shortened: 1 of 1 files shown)


R/system.R
```diff
@@ -490,7 +490,7 @@ clearAmazonAthenaConnection <- function(driver = "", region = "", authentication
 #' If not, new connection is created and returned.
 #' @export
 getDBConnection <- function(type, host = NULL, port = "", databaseName = "", username = "", password = "", catalog = "", schema = "", dsn="", additionalParams = "",
-                            collection = "", isSSL = FALSE, authSource = NULL, cluster = NULL, timeout = NULL, connectionString = NULL) {
+                            collection = "", isSSL = FALSE, authSource = NULL, cluster = NULL, timeout = NULL, connectionString = NULL, driver = "") {
```
hideaki: Why is the default for driver set to "" rather than NA?

R/system.R
```diff
@@ -705,6 +705,50 @@ getDBConnection <- function(type, host = NULL, port = "", databaseName = "", use
         connection_pool[[key]] <- conn
       }
     }
+  } else if (type == "mssqlserver") {# The type sqlserver is already used for RODBC based one and "mssqlserver" is passed from Exploratory Desktop.
+
+    # if platform is Linux use predefined one
```
hideaki: Please add comment on the reason of this.

R/system.R
```diff
@@ -991,19 +1039,30 @@ queryAmazonAthena <- function(driver = "", region = "", authenticationType = "IA
 #' @param port - Database port number
 #' @param as.is - Flag to tell if you honor data types from ODBC
 #'
-queryODBC <- function(dsn,username, password, additionalParams, numOfRows = 0, query, stringsAsFactors = FALSE, host="", port="", as.is = TRUE, ...){
-  if(!requireNamespace("RODBC")){stop("package RODBC must be installed.")}
-
-  conn <- getDBConnection("odbc", host, port, NULL, username, password, dsn = dsn, additionalParams = additionalParams)
+queryODBC <- function(dsn="", username, password, additionalParams="", numOfRows = 0, query, stringsAsFactors = FALSE, host="", port="", as.is = TRUE, databaseName="", driver = "", type = "", ...){
```
hideaki: Please explain when type can be other than "odbc" here in the comment.

R/system.R
```diff
@@ -1017,9 +1076,12 @@ queryODBC <- function(dsn,username, password, additionalParams, numOfRows = 0, q
   }, error = function(err) {
     # for some cases like conn not being an open connection, sqlQuery still throws error. handle it here.
     # clear connection in pool so that new connection will be used for the next try
-    clearDBConnection("odbc", NULL, NULL, NULL, username, dsn = dsn, additionalParams = additionalParams)
+    clearDBConnection(type, NULL, NULL, NULL, username, dsn = dsn, additionalParams = additionalParams)
     stop(err)
   })
+  if(type == "mssqlserver") {
+    DBI::dbClearResult(resultSet)
```
hideaki: Please explain why this is necessary only for mssqlserver case, if that is the case.

R/system.R
```diff
@@ -705,6 +705,50 @@ getDBConnection <- function(type, host = NULL, port = "", databaseName = "", use
...
       }
     }
+  } else if (type == "mssqlserver") {# The type sqlserver is already used for RODBC based one and "mssqlserver" is passed from Exploratory Desktop.
+
+    # if platform is Linux use predefined one
+    if(Sys.info()["sysname"]=="Linux"){
+      driver <-  "ODBC Driver 17 for SQL Server";
+    }
+    if(!requireNamespace("DBI")){stop("package DBI must be installed.")}
+    if(!requireNamespace("odbc")){stop("package RPostgres must be installed.")}
```
hideaki: Please add comment on why we are using odbc instead of RODBC here.

hidekoji:
Tested connection failure and fallback

```r
Error in connection_sql_tables(conn@ptr, catalog_name = catalog_name,  :
  nanodbc/nanodbc.cpp:4357: 00000: [Microsoft][ODBC Driver 17 for SQL Server]TCP Provider: Error code 0x274C  [Microsoft][ODBC Driver 17 for SQL Server]Communication link failure 
```

And I can get a valid connection after that.

R/system.R
hidekoji: Changed it to NULL

R/system.R
hidekoji: Added comment.

R/system.R
hidekoji: Added comment.

R/system.R
hidekoji: Added comment.

R/system.R
hidekoji: Added comment.

PR #784, opened 2020-05-12
Cox survival curve PDP conf int value fix

hideaki:
# Description
- Cox survival curve PDP conf int fix
- Apply grid to PDP survival curves
- Adjust PDP sample size for cox, survival forest

# Checklist

- [ ] Pass devtools::check()
- [ ] Pass devtools::test()
- [ ] Tested with Exploratory


R/build_coxph.R
```diff
@@ -94,7 +94,10 @@ partial_dependence.coxph_exploratory <- function(fit, time_col, vars = colnames(
   }
 
   aggregate.fun <- function(x) {
-    mean(x, na.rm = TRUE)
+    # na.rm needs to be FALSE, since even when only one of the sample has
+    # NA confidende interval, the mean conf int of the rest would not cleanly match with the mean
```
hideaki, line 98:
```suggestion
    # NA confidence interval, the mean conf int of the rest would not cleanly match with the mean
```

kei51e:
Code change looks good.

PR #799, opened 2020-06-06
fix Data Source: MySQL: Changed default from integer64 to numeric for bigint data mapping

hidekoji:
# Description

For MySQL and other DBI based data sources, change it to pass "numeric" for `bigint` argument.

# Checklist

- [ ] Pass devtools::check()
- [ ] Pass devtools::test()
- [x] Tested with Exploratory

### DIFF
--- a/R/system.R
+++ b/R/system.R
@@ -573,7 +573,7 @@ getDBConnection <- function(type, host = NULL, port = "", databaseName = "", use
     }
     # if the connection is null or the connection is invalid, create a new one.
     if (is.null(conn) || !DBI::dbIsValid(conn)) {
-      # To avoid integer64 issue, pass numeric for bigint argument.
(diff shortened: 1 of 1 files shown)


R/system.R
```diff
@@ -573,8 +573,9 @@ getDBConnection <- function(type, host = NULL, port = "", databaseName = "", use
     }
     # if the connection is null or the connection is invalid, create a new one.
     if (is.null(conn) || !DBI::dbIsValid(conn)) {
+      # To avoid integer64 issue, pass numeric for bigint argument.
```
hideaki, line 576:
```suggestion
      # To avoid integer64 handling issues in charts, etc., use numeric as the R type to receive bigint data rather than default integer64 by specifying bigint argument.
```

PR #804, opened 2020-06-17
fix issue that passing non-ascii character to parse number causes an issue.

hidekoji:
# Description

Fix the issue that passing non-ASCII character to parse number causes : `Error in nchar(x): invalid multibyte string, element 4`

# Checklist

- [ ] Pass devtools::check()
- [ ] Pass devtools::test()
- [x] Tested with Exploratory

### DIFF
--- a/R/string_operation.R
+++ b/R/string_operation.R
@@ -555,14 +555,14 @@ parse_number <- function(text, ...){
     # so explicitly convert it to character before calling readr::parse_number
     # Passing non-ascii character to readr::parse_number causes an error so remove non-ascii character before calling readr::parse_number.
     # ref: https://github.com/tidyverse/readr/issues/1111
-    as.numeric(readr::parse_number(gsub('[^ -~]', '', as.character(text)), ...))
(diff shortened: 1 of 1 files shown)


R/string_operation.R
```diff
@@ -553,15 +553,20 @@ parse_number <- function(text, ...){
   } else if (!is.character(text)) {
     # non character data raises Error in parse_vector(x, col_number(), na = na, locale = locale, trim_ws = trim_ws) : is.character(x) is not TRUE
     # so explicitly convert it to character before calling readr::parse_number
-    as.numeric(readr::parse_number(as.character(text), ...))
+    # Passing non-ascii character causes an error so, remove non-ascii character before calling readr::parse_number.
+    # ref: https://github.com/tidyverse/readr/issues/1111
+    as.numeric(readr::parse_number(gsub('[^ -~]', '', as.character(text)), ...))
```
kei51e, line 558: Should be better to use [^\x20-\x7E] instead. It is hard to tell "[^ -~]" means non-ascii range .

hidekoji: Thank you for the suggestion. I updated accordingly.

PR #809, opened 2020-06-22
Use mutate/purrr::map instead of rowwise tidier for prediction_survfit

hideaki:
# Description
- Use mutate/purrr::map instead of rowwise tidier for prediction_survfit
- Add necessary ungroup() in model_confint()

# Checklist

- [ ] Pass devtools::check()
- [ ] Pass devtools::test()
- [ ] Tested with Exploratory


hideaki:
The errors are the ones which will fix when we replace the test docker image.

hidekoji:

PR #849, opened 2020-08-29
XGBoost Analytics View function

hideaki:
# Description
- XGBoost Analytics View function
- New data prediction with Analytics View models.

# Checklist

- [ ] Pass devtools::check()
- [ ] Pass devtools::test()
- [x] Tested with Exploratory


R/broom_wrapper.R
```diff
@@ -508,10 +512,20 @@ prediction <- function(df, data = "training", data_frame = NULL, conf_int = 0.95
 
       # Use formula to support expanded aug_args (especially for type.predict for logistic regression)
       # because ... can't be passed to a function inside mutate directly.
-      aug_fml <- if(aug_args == ""){
-        "broom::augment(m, data = df)"
-      } else {
-        paste0("broom::augment(m, data = df, ", aug_args, ")")
+      if (!is.null(df$model[[1]]$prediction_training) || # Check if model already has training prediction result. # TODO: Make this model agnostic.
```
kei51e, line 515: Is `df$model[[1]` itself guaranteed not null?

Is "lm_exploratory" (or "glm_exploratory") guaranteed to show up always at the beginning of the vector?

R/build_xgboost.R
```diff
@@ -277,69 +289,107 @@ xgboost_reg <- function(data, formula, output_type = "linear", eval_metric = "rm
...
+      cleaned_data <- align_predictor_factor_levels(cleaned_data, x$df, predictor_variables)
+    }
 
-    vars <- all.vars(x$terms)
-    y_name <- vars[[1]]
+    # For new data prediction, xgboost can predict with NAs in the predictors. Disable NA filtering for now.
+    # na_row_numbers <- ranger.find_na(predictor_variables, cleaned_data)
+    na_row_numbers <- c()
+    if (length(na_row_numbers) > 0) {
+      cleaned_data <- cleaned_data[-na_row_numbers,]
```
kei51e, line 323: You may want to remove or comment out this part.

R/build_xgboost.R
```diff
@@ -629,3 +745,627 @@ glance.xgb.Booster <- function(x, pretty.name = FALSE, ...) {
...
+    # whose number of rows is the same with its size,
+    # so the result should be filled by NA
+    predicted <- fill_vec_NA(as.integer(rownames(mat_data)), predicted, nrow(df))
+  }
+  predicted
+}
+
+
+partial_dependence.xgboost <- function(fit, vars = colnames(data),
+  n = c(min(nrow(unique(data[, vars, drop = FALSE])), 25L), nrow(data)),
```
kei51e, line 795: How can I change 25? Probably we should have another argument for max number of categories?

R/build_xgboost.R
kei51e: Actually I think this is ok after reading the code.

R/randomForest_tidiers.R
```diff
@@ -1015,18 +1041,70 @@ augment.rpart <- function(x, data = NULL, newdata = NULL, ...) {
...
-    newdata[predicted_value_col] <- predicted_value
-    newdata[predicted_probability_col] <- apply(predict(x, newdata), 1, max)
-    newdata %>% dplyr::rename(predicted_label = predicted_value_col) %>%
-                dplyr::select(-predicted_label, everything(), predicted_label)
+
+    if (x$classification_type == "binary") {
+      newdata[[predicted_value_col]] <- predicted_value
+
+      # With ranger, 2nd category always is the one to be considered "TRUE",
+      predicted_prob <- pred_res[, 2]
```
kei51e, line 1092: Better to use the column name instead of an index if the column name is stable.

R/broom_wrapper.R
hideaki: The expression works even if df$model[[1]] is NULL.
lm/glm_exploratory class is always set at the beginning of the vector.

R/randomForest_tidiers.R
hideaki: Made that change.

R/broom_wrapper.R
kei51e: Thanks for your clarification. I confirmed it.

R/randomForest_tidiers.R
kei51e: Thanks.

PR #870, opened 2020-10-12
Make %in_or_all% work for our case-insensitive filter condition.

hideaki:
# Description
Make %in_or_all% work for our case-insensitive filter condition.

# Checklist

- [ ] Pass devtools::check()
- [ ] Pass devtools::test()
- [x] Tested with Exploratory


hidekoji (approved):
It looks good to me.

PR #885, opened 2020-11-16
Data Wrangling: Support an option to summarize the count by Token, not by Document ID

hidekoji:
# Description

As for do_tokenize_icu, change it to show summarized count by token if drop argument is TRUE and keep_cols argument is FALSE.

# Checklist

- [ ] Pass devtools::check()
- [ ] Pass devtools::test()
- [x] Tested with Exploratory

### DIFF
--- a/R/string_operation.R
+++ b/R/string_operation.R
@@ -173,7 +173,7 @@ do_tokenize_icu <- function(df, text_col, token = "word", keep_cols = FALSE,
                                  remove_symbols = TRUE, remove_twitter = TRUE,
                                  remove_url = TRUE, stopwords_lang = NULL,
                                  hiragana_word_length_to_remove = 2,
-                                 summary_level = "row", sort_by = "count", ...){
+                                 summary_level = "row", sort_by = "", ...){
 
   if(!requireNamespace("quanteda")){stop("package quanteda must be installed.")}
   if(!requireNamespace("dplyr")){stop("package dplyr must be installed.")}
--- a/R/string_operation.R
+++ b/R/string_operation.R
@@ -250,7 +250,7 @@ do_tokenize_icu <- function(df, text_col, token = "word", keep_cols = FALSE,
   }
   # if the summary_level is "all", summarize it by token.
   if(summary_level == "all") {
-    result <- result %>% dplyr::group_by(!!rlang::sym(token_col)) %>% dplyr::summarise(!!rlang::sym(count_col) := sum(!!rlang::sym(count_col)))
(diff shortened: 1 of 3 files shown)


R/string_operation.R
```diff
@@ -241,7 +244,16 @@ do_tokenize_icu <- function(df, text_col, token = "word", keep_cols = FALSE,
   }
   # result column order should be document_id, <token_col>, <count_col> ...
   if(!with_id) {
-    result %>% select(!!rlang::sym(token_col), !!rlang::sym(count_col), dplyr::everything())
+    result <-result %>% select(!!rlang::sym(token_col), !!rlang::sym(count_col), dplyr::everything())
+    if(drop && !keep_cols && summary_level == "token") { # if the summary_level is token and resulting columns are only token and count, summarize it by token.
+      result <- result %>% dplyr::group_by(!!rlang::sym(token_col)) %>% dplyr::summarise(!!rlang::sym(count_col) := n())
```
hideaki, line 249: This should be sum rather than n().

R/string_operation.R
```diff
@@ -241,7 +244,16 @@ do_tokenize_icu <- function(df, text_col, token = "word", keep_cols = FALSE,
   }
   # result column order should be document_id, <token_col>, <count_col> ...
   if(!with_id) {
-    result %>% select(!!rlang::sym(token_col), !!rlang::sym(count_col), dplyr::everything())
+    result <-result %>% select(!!rlang::sym(token_col), !!rlang::sym(count_col), dplyr::everything())
+    if(drop && !keep_cols && summary_level == "token") { # if the summary_level is token and resulting columns are only token and count, summarize it by token.
```
hideaki, line 248: Conditions other than summary_level seems unnecessary.

R/string_operation.R
```diff
@@ -170,7 +172,8 @@ do_tokenize_icu <- function(df, text_col, token = "word", keep_cols = FALSE,
                                  remove_hyphens = FALSE, remove_separators = TRUE,
                                  remove_symbols = TRUE, remove_twitter = TRUE,
                                  remove_url = TRUE, stopwords_lang = NULL,
-                                 hiragana_word_length_to_remove = 2, ...){
+                                 hiragana_word_length_to_remove = 2,
+                                 summary_level = "document", sort_by = "count", ...){
```
hideaki, line 176: summary_level:document/token should be all/row.

R/string_operation.R
hidekoji: fixed it to use sum() instead of n()

R/string_operation.R

R/string_operation.R
hidekoji: updated as suggested.

PR #889, opened 2020-11-20
On Windows, Web Scraping shows invalid multibyte error for some of the Japanese characters.

hidekoji:
# Description

On Windows, Web Scraping shows invalid multibyte error for some of the Japanese characters.
To address the issue, change the code to switch local to "C" before data import then bring back the original locale once it's done.

# Checklist

- [ ] Pass devtools::check()
- [ ] Pass devtools::test()
- [x] Tested with Exploratory

### DIFF
--- a/R/system.R
+++ b/R/system.R
@@ -1616,8 +1616,18 @@ parse_html_tables <- function(url, encoding = NULL) {
 scrape_html_table <- function(url, index, heading, encoding = NULL) {
   loadNamespace("rvest");
   loadNamespace("tibble");
+  current_locale = Sys.getlocale(category = "LC_CTYPE")
   .htmltables <- parse_html_tables(url, encoding)
-  tibble::repair_names(rvest::html_table(.htmltables[[index]], fill=TRUE ,header=heading))
+  if (Sys.info()["sysname"] == "Windows") {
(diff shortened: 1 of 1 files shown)


R/system.R
```diff
@@ -1616,8 +1616,18 @@ parse_html_tables <- function(url, encoding = NULL) {
...
   loadNamespace("tibble");
+  current_locale = Sys.getlocale(category = "LC_CTYPE")
   .htmltables <- parse_html_tables(url, encoding)
-  tibble::repair_names(rvest::html_table(.htmltables[[index]], fill=TRUE ,header=heading))
+  if (Sys.info()["sysname"] == "Windows") {
+    # For Windows, temporary change LC_CTYPE to C to workaround the "invalid multibyte string" error raised from rvest.
+    Sys.setlocale(category="LC_CTYPE", locale="C");
+  }
+  res <- tibble::repair_names(rvest::html_table(.htmltables[[index]], fill=TRUE ,header=heading))
+  if (Sys.info()["sysname"] == "Windows") {
```
hideaki, line 1626: Use tryCatch to make sure locale gets back to original even in error case.

hidekoji: Changed to use tryCacth to make sure locale gets back to the original one even with the error case.

PR #909, opened 2020-12-14
Summarize Group: Change resulting data type from integer to numeric.

hidekoji:
# Description

For Summarize Group, change the resulting data type from integer to numeric.

# Checklist

- [ ] Pass devtools::check()
- [ ] Pass devtools::test()
- [x] Tested with Exploratory

### DIFF
--- a/R/util.R
+++ b/R/util.R
@@ -2244,7 +2244,8 @@ summarize_group <- function(.data, group_cols = NULL, group_funs = NULL, ...){
       }
     }
   }
-  ret
+  # For integer columns (like # of rows, unique) change them to numeric for better usability.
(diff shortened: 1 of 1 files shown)


R/util.R
```diff
@@ -2244,7 +2244,8 @@ summarize_group <- function(.data, group_cols = NULL, group_funs = NULL, ...){
       }
     }
   }
-  ret
+  # For integer columns (like # of rows, unique) change them to numeric for better usability.
```
hideaki, line 2247: Can you leave comment on why we do it?

hidekoji: Added comment with an example of why we want to do it.

PR #918, opened 2021-01-09
exp_ts_cluster update

hideaki:
# Description
- Fill left/right data too.
- Aggregate variables that are not used for clustering but want to include in the output
- Remove categories that do not have enough data from the input.

# Checklist

- [ ] Pass devtools::check()
- [ ] Pass devtools::test()
- [x] Tested with Exploratory


R/na_util.R
```diff
@@ -158,32 +158,48 @@ fill_between <- function(df, ..., .direction="down", value=NULL) {
 #'                       "spline" - Spline interpolation.
 #'                       NULL - Skip NA fill. Use this only when you know there is no NA.
 #' @param val - Value to fill NA when na_fill_type is "value"
-fill_ts_na <- function(target, time, type = "previous", val = 0) {
+fill_ts_na <- function(target, time, type = c("value", "previous", "extend"), val = c(0, 0, 0)) {
+  if (length(type) == 1) {
```
kei51e, line 162: Add a comment for type and value structure.

PR #919, opened 2021-01-13
Updated getDBConnection and queryODBC to support snowflake

hidekoji:
# Description

Updated getDBConnection and queryODBC to support snowflake.

It seems the Unit test fails with exp_ts_cluster which has nothing to the PR.

# Checklist

- [ ] Pass devtools::check()
- [ ] Pass devtools::test()
- [x] Tested with Exploratory

### DIFF
--- a/R/system.R
+++ b/R/system.R
@@ -677,7 +677,7 @@ clearAmazonAthenaConnection <- function(driver = "", region = "", authentication
 #' If not, new connection is created and returned.
 #' @export
 getDBConnection <- function(type, host = NULL, port = "", databaseName = "", username = "", password = "", catalog = "", schema = "", dsn="", additionalParams = "",
-                            collection = "", isSSL = FALSE, authSource = NULL, cluster = NULL, timeout = NULL, connectionString = NULL, driver = NULL) {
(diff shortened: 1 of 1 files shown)


R/system.R
```diff
@@ -1029,6 +1029,61 @@ getDBConnection <- function(type, host = NULL, port = "", databaseName = "", use
                              bigint = "numeric")
       connection_pool[[key]] <- conn
     }
+  } else if (type == "snowflake") {
+    # If the platform is Linux, set the below predefined driver installed on Collaboration Server
+    # so that this data source can be scheduled.
```
hideaki, line 1034: Add comment on how this works - it overwrites the value for desktop when run on server.

R/system.R
```diff
@@ -1113,7 +1168,7 @@ clearDBConnection <- function(type, host = NULL, port = NULL, databaseName, user
 }
 
 isConnecitonPoolEnabbled <- function(type){
```
hideaki, line 1170: Correct typo.

R/system.R
hidekoji: Added comment.

R/system.R
hidekoji: fixed typo.

PR #937, opened 2021-02-07
Keep column order after add_prediction for Cox regression, survival forest, lm, glm

hideaki:
# Description
Keep column order for add_prediction for Cox regression, survival forest, lm, glm.

# Checklist

- [ ] Pass devtools::check()
- [ ] Pass devtools::test()
- [x] Tested with Exploratory


hidekoji (approved):
it looks good to me.

PR #949, opened 2021-03-09
fix: do_ngram does not work if document_id and sentence_id are same column.

hidekoji:
# Description

fix issue that `do_ngram` does not work if `dcoument_id` and `sentence_id` are the same column.

# Checklist

- [ ] Pass devtools::check()
- [ ] Pass devtools::test()
- [x] Tested with Exploratory


R/string_operation.R
```diff
@@ -510,8 +510,14 @@ do_ngram <- function(df, token, sentence, document, maxn=2, sep="_"){
...
   # this is executed for ngrams not to be connected over sentences
-  grouped <- df %>%
-    dplyr::group_by(!!!rlang::syms(c(document_col, sentence_col))) # convert the column name to symbol for colum names with backticks
+  grouped <- NULL
+  if (sentence_col != document_col) {
+    grouped <- df %>%
+      dplyr::group_by(!!!rlang::syms(c(document_col, sentence_col))) # convert the column name to symbol for column names with backticks
+  } else { # if document_col and sentence_col are identical, just group the data frame by the document_col.If you group by with identical columns, group_by throws an error.
+    grouped <- df %>%
+      dplyr::group_by(!!!rlang::syms(c(document_col))) # convert the column name to symbol for column names with backticks
```
hideaki, line 519: c() does not seem necessary.
rlang::sym seems to be enough since it is handling a single column.

R/string_operation.R
```diff
@@ -510,8 +510,14 @@ do_ngram <- function(df, token, sentence, document, maxn=2, sep="_"){
   document_col <- col_name(substitute(document))
 
   # this is executed for ngrams not to be connected over sentences
-  grouped <- df %>%
-    dplyr::group_by(!!!rlang::syms(c(document_col, sentence_col))) # convert the column name to symbol for colum names with backticks
+  grouped <- NULL
+  if (sentence_col != document_col) {
+    grouped <- df %>%
+      dplyr::group_by(!!!rlang::syms(c(document_col, sentence_col))) # convert the column name to symbol for column names with backticks
+  } else { # if document_col and sentence_col are identical, just group the data frame by the document_col.If you group by with identical columns, group_by throws an error.
```
hideaki, line 517: Please format this comment better.

R/string_operation.R
hidekoji: Changed it to use ```!!rlang::sym``` as suggested.

R/string_operation.R
hidekoji: format the comment.

PR #968, opened 2021-04-07
Parameter: Handle empty vector with class such as as.character(c()) correctly

hideaki:
# Description
Handle empty vector with class such as as.character(c()) correctly as a parameter value in generating SQL/MongoDB query.

# Checklist

- [ ] Pass devtools::check()
- [ ] Pass devtools::test()
- [x] Tested with Exploratory

### DIFF
--- a/R/system.R
+++ b/R/system.R
@@ -339,7 +339,7 @@ js_glue_transformer <- function(expr, envir) {
     }
   }
 
-  if (is.null(val)) { # NULL in R is same as empty vector. Print empty string.
+  if (length(val) == 0) { # Empty vector case. NULL in R is same as empty vector, but since is.null(as.character(c(0))) returns FALSE, this has to be done my length().
     val <- ""
   }
   else if (is.numeric(val)) {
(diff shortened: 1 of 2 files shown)


R/system.R
```diff
@@ -339,7 +339,7 @@ js_glue_transformer <- function(expr, envir) {
     }
   }
 
-  if (is.null(val)) { # NULL in R is same as empty vector. Print empty string.
+  if (length(val) == 0) { # Empty vector case. NULL in R is same as empty vector, but since is.null(as.character(c(0))) returns FALSE, this has to be done my length().
```
hidekoji, line 342: Do you mean "by length()" by "my length()"?

R/system.R
```diff
@@ -469,7 +469,8 @@ sql_glue_transformer_internal <- function(expr, envir, bigquery=FALSE) {
...
```
hidekoji, line 472: Do you mean "by length()" by "my length()"?

R/system.R
hideaki: That's right. Fixed it. Thanks!

R/system.R
hideaki: Fixed it. Thanks!

hideaki:
The jenkins test errors are the expected ones until we update test docker image.

PR #984, opened 2021-05-04
Test: Add test to detect arrow::read_parquet regression in arrow3.

kei51e:
# Description
Add test to detect arrow::read_parquet regression in arrow3. 

# Checklist

- [ ] Pass devtools::check()
- [x] Pass devtools::test()
- [ ] Tested with Exploratory


hidekoji (approved):
it looks good to me.

PR #1012, opened 2021-06-13
Handle numeric column/row name case in mat_to_df

hideaki:
# Description
In mat_to_df, making sure to return character column/row names in the long-style output data frame even when the input matrix has numeric column/row names.

# Checklist

- [ ] Pass devtools::check()
- [ ] Pass devtools::test()
- [x] Tested with Exploratory


hidekoji (approved):

PR #1014, opened 2021-06-15
Tokenize specified compound word so that user does not need to separate words in it

hideaki:
# Description
Tokenize specified compound word so that user does not need to separate words in it.

# Checklist

- [ ] Pass devtools::check()
- [ ] Pass devtools::test()
- [x] Tested with Exploratory


hidekoji (approved):

PR #1026, opened 2021-07-06
Delete if "Centroid <number>" is within the groups of input data

hideaki:
# Description
Delete if "Centroid <number>" is within the groups of input data, since they would get mixed up with real centroid we are going to calculate.

# Checklist

- [ ] Pass devtools::check()
- [ ] Pass devtools::test()
- [x] Tested with Exploratory


hidekoji (approved):
it looks good to me.

PR #1058, opened 2021-08-14
Support load_fred API

hidekoji:
# Description

Support load_fred

# Checklist

- [ ] Pass devtools::check()
- [ ] Pass devtools::test()
- [x] Tested with Exploratory


R/system.R
```diff
@@ -2953,3 +2953,21 @@ filter_cascade <- function(.data, ...){
   }
   df
 }
+
+#
+#'@export
+load_fred <- function(series_id, date_start = "", date_end = NULL, password) {
```
hideaki, line 2959: Why is date_start default is empty string while the date_end default is NULL?

R/system.R
```diff
@@ -2953,3 +2953,21 @@ filter_cascade <- function(.data, ...){
   }
   df
 }
+
+#
+#'@export
```
hideaki, line 2958: Add some comment to describe this function.

R/system.R
hidekoji: The default for date_end should be an empty string so I updated the default as "".

R/system.R
hidekoji: Added comment.

PR #1061, opened 2021-08-16
fix str_remove_word fails when a word contains special characters.

hidekoji:
# Description

str_remove_word fails if the text is something like `"Great White Multi-Use Recycled Paper (20Lb. and 84 Bright)` and word is created with space as a separator.

This is because it ended up doing below incorrect reg exp operation.

```r
stringr::str_replace("Great White Multi-Use Recycled Paper (20Lb. and 84 Bright)", stringr::str_c(" ", "Bright)", "$"), "")
```

# Checklist

- [ ] Pass devtools::check()
- [ ] Pass devtools::test()
- [x] Tested with Exploratory

### DIFF
--- a/R/string_operation.R
+++ b/R/string_operation.R
@@ -837,26 +837,26 @@ str_replace_word <- function(string, start = 1L, end = start, sep = fixed(" "), 
     if(end == 1) { # It means the first word.
       if (rep == "") { # for remove case
         x[1] = rep # replace first word with the replace text.
-        if (len > 1) { # trim the white space of the second element's left side
+        if (len > 1) { # trim the white space of the second word's left side
           x[2] = stringr::str_trim(x[2], side = "left")
         }
-        # join back the string using the normalized separator.
+        # join back the words using the normalized separator.
         stringr::str_c(x[-1], collapse = sep_)
       } else {
-        x[1] = rep
+        x[1] = rep # replace the first word with the replace text.
         stringr::str_c(x, collapse = sep_)
       }
     } else if (end == -1){ # It means the last word.
       if (rep == "") { # for Remove Case
-        if (len > 1) {# trim the white space of the second element from the end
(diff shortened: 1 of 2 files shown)


tests/testthat/test_string_operation.R
```diff
@@ -666,6 +666,9 @@ test_that("str_remove_word", {
   expect_equal(ret, c("Sequoia Capital China, Qiming Venture Partners"))
   ret <- exploratory::str_remove_word("Sequoia Capital China, Qiming Venture Partners, Tencent Holdings", 1, sep = "\\s*\\,\\s*")
   expect_equal(ret, c("Qiming Venture Partners, Tencent Holdings"))
+  ret <- exploratory::str_remove_word("Sequoia Capital (China Space), Qiming Venture Partners, Tencent Holdings", 1, sep = "\\s*\\,\\s*")
```
hideaki, line 669: Correct the test case.

hidekoji: fixed the test case.

R/string_operation.R
```diff
@@ -831,24 +831,39 @@ str_replace_word <- function(string, start = 1L, end = start, sep = fixed(" "),
...
+    } else if (end == -1){ # It means the last word.
+      if (rep == "") { # for Remove Case
+        if (len > 1) {# trim the white space of the second element from the end
+          x[len - 1] = stringr::str_trim(x[len - 1], side = "right")
+        }
+        # drop the last element from list then collapse with the separator.
+        stringr::str_c(x[-1 * len], collapse = sep_)
+      } else {
+        # replace the last element with the replace string.
+        x[len] = rep
```
hideaki, line 858: If the sep_ is ",", we could prefix rep with " ", so that the result looks prettier. (The test result would be the same as before.)

hideaki:
The error is unrelated data source test issue.

PR #1066, opened 2021-08-19
Fix FRED start date handling: Change date_start parameter as an optional parameter

hidekoji:
# Description

Change date_start parameter as an optional parameter.

# Checklist

- [ ] Pass devtools::check()
- [ ] Pass devtools::test()
- [x] Tested with Exploratory

### DIFF
--- a/R/system.R
+++ b/R/system.R
@@ -2968,8 +2968,7 @@ filter_cascade <- function(.data, ...){
 
 #'API to load economic data from FRED (Federal Reserve Bank Economic Data)
 #'@param series_id - e.g. UNRATE
-#'@param date_start - Start Date for the query. This is required field.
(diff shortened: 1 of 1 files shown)


R/system.R
```diff
@@ -2981,10 +2981,18 @@ load_fred <- function(series_id, date_start = "", date_end = "", password) {
   } else {
     date_end <- lubridate::ymd(date_end)
   }
-  fredr::fredr(
-    series_id = series_id,
-    observation_start = lubridate::ymd(date_start),
-    observation_end = date_end
-  )
+  # date_start is an optional parameter, so if it's not specified, execute the query without the start_date.
```
hideaki, line 2984: Update the comment about date_start.
```
#'@param date_start - Start Date for the query. This is required field.
#'                    When it's called from Exploratory Desktop, this date_start is always passed since it has required input field validation.
```

PR #1140, opened 2022-01-27
Use base::as.pairlist instead of removed rlang::as_pairlist

hideaki:
# Description
Use base::as.pairlist instead of removed experimental rlang::as_pairlist.

# Checklist

- [ ] Pass devtools::check()
- [ ] Pass devtools::test()
- [x] Tested with Exploratory

### DIFF
--- a/R/util.R
+++ b/R/util.R
@@ -1375,7 +1375,7 @@ do_on_each_group <- function(df, func, params = quote(list()), name = "tmp", wit
   name <- avoid_conflict(colnames(df), name)
   # This is a list of arguments in do clause
   args <- append(list(quote(.)), rlang::call_args(params))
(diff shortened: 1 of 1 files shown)


R/util.R
```diff
@@ -1375,7 +1375,7 @@ do_on_each_group <- function(df, func, params = quote(list()), name = "tmp", wit
   name <- avoid_conflict(colnames(df), name)
   # This is a list of arguments in do clause
   args <- append(list(quote(.)), rlang::call_args(params))
-  call <- rlang::new_call(func, rlang::as_pairlist(args))
+  call <- rlang::new_call(func, as.pairlist(args))
```
kei51e, line 1378: Add either `base::`, or if it is too much, add a comment that we explicitly use `base::as.pairlist` to avoid using `rlang::as_pairlist`.

PR #1161, opened 2022-02-17
is_japanese_holiday: Workaround for NA

hidekoji:
# Description

Added a workaround for is_japanse_holiday fails when NA is passed.

# Checklist

- [ ] Pass devtools::check()
- [ ] Pass devtools::test()
- [x] Tested with Exploratory

### DIFF
--- a/R/util.R
+++ b/R/util.R
@@ -1634,8 +1634,8 @@ is_japanese_holiday <- function(date) {
     date <-
       lubridate::as_date(date)
       # make sure to exclude NA otherwise, lubridate::as_date(unlist(zipangu::jholiday(yr, "en")))
-      yr <-
-        unique(lubridate::year(date[!is.na(date)]))
+    yr <-
+      unique(lubridate::year(date[!is.na(date)]))
     jholidays <-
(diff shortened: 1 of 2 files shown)


R/util.R
```diff
@@ -1624,8 +1624,31 @@ weekend <- function(x){
 is_japanese_holiday <- function(date) {
   current_option <- getOption("lubridate.week.start")
   result <- tryCatch({
+    # Make sure to set 7 as the week start date to make the result stable.
```
hideaki, line 1627: Explain some more detail.

R/util.R
```diff
@@ -1624,8 +1624,31 @@ weekend <- function(x){
...
     options(lubridate.week.start = 7)
-    zipangu::is_jholiday(date)
+    date <-
+      lubridate::as_date(date)
+    na_index <- which(sapply(date, is.na))
```
hideaki, line 1631: is.na(date) would be enough for this purpose.

R/util.R
hidekoji: Added comment.

PR #1180, opened 2022-03-27
When data type override with date is specifically set, make sure to convert POSIXct to Date.

hidekoji:
# Description

When data type override with date is specifically set, make sure to convert POSIXct to Date.

# Checklist

- [ ] Pass devtools::check()
- [ ] Pass devtools::test()
- [x] Tested with Exploratory

### DIFF
--- a/R/system.R
+++ b/R/system.R
@@ -2824,7 +2824,13 @@ read_excel_file <- function(path, sheet = 1, col_names = TRUE, col_types = NULL,
     }
     if(!is.null(tzone)) { # if timezone is specified, apply the timezeon to POSIXct columns
       df <- df %>% dplyr::mutate_if(lubridate::is.POSIXct, funs(lubridate::force_tz(., tzone=tzone)))
+      # make sure to convert POSIXct to Date if the col_types are set as Date specifically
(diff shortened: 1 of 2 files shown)


R/system.R
```diff
@@ -2824,7 +2824,13 @@ read_excel_file <- function(path, sheet = 1, col_names = TRUE, col_types = NULL,
     }
     if(!is.null(tzone)) { # if timezone is specified, apply the timezeon to POSIXct columns
       df <- df %>% dplyr::mutate_if(lubridate::is.POSIXct, funs(lubridate::force_tz(., tzone=tzone)))
+      # make sure to convert POSIXct to Date if the col_types are set as Date specifically
```
hideaki, line 2827: Explain the behavior of readxl.

R/system.R
```diff
@@ -2824,7 +2824,13 @@ read_excel_file <- function(path, sheet = 1, col_names = TRUE, col_types = NULL,
...
+      df <- df %>% dplyr::mutate_at(vars(colnames(df)[col_types == "date"]), funs(as.Date(.,tz=tzone)))
```
hideaki, line 2828: Look into the behavior of readxl, and make sure setting timezone would not make the result different from the original date.

PR #1183, opened 2022-03-29
Handle errors in e$parent that seems to come with newer dplyr, etc.

hideaki:
# Description
Handle errors in e$parent that seems to come with newer dplyr, etc.

# Checklist

- [ ] Pass devtools::check()
- [ ] Pass devtools::test()
- [ ] Tested with Exploratory


hidekoji:

PR #1233, opened 2022-08-02
Support target_columns arguments for join to support column selection as preprocess

hidekoji:
# Description

a) Supported source_column, target_columns argument for join command.

- When source_columns are provided, it first only selects the columns from the source data frame and then joins data frames.
- When target_columns are provided, it first only selects the columns from the target data frame and then joins data frames.

b) Fixed suffix does not work for insensitve_join

# Checklist

- [ ] Pass devtools::check()
- [ ] Pass devtools::test()
- [x] Tested with Exploratory

### DIFF
--- a/R/join.R
+++ b/R/join.R
@@ -24,9 +24,11 @@ insensitive_join <- function(fun = dplyr::left_join, type = "LEFT") {
 #' Wrapper function for dplyr's inner_join to support case insensitive join.
 #' @export
 inner_join <- function(x, y, by = NULL, copy = FALSE, suffix = c(".x", ".y"), ignorecase = FALSE, source_columns = NULL, target_columns = NULL, ...){
+  # Limit source columns to use for join when source_columns are set.
   if (!is.null(source_columns)) {
     x <- x %>% dplyr::select(source_columns)
   }
+  # Limit target columns to use for join when target_columns are set.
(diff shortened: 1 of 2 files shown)


kei51e (approved):
Code change looks good to me.

R/join.R
```diff
@@ -4,26 +4,30 @@ insensitive_join <- function(fun = dplyr::left_join, type = "LEFT") {
...
     tmp_by_y <- paste0("_", by$y, "_")
     for (i in seq_along(by$x)) {
       x[[tmp_by_x[[i]]]] <- tolower(x[[by$x[[i]]]])
       y[[tmp_by_y[[i]]]] <- tolower(y[[by$y[[i]]]])
       y[[by$y[[i]]]] <- NULL
     }
-    
-    res <- fun(x, y, list(x = tmp_by_x, y = tmp_by_y))
+
+    res <- fun(x, y, list(x = tmp_by_x, y = tmp_by_y), suffix = suffix)
```
hideaki, line 16: suffix seems to be undefined in this function.

hideaki: Also, shouldn't the copy argument passed down too?

hidekoji: Add comment for for suffix and copy arguments.

PR #1239, opened 2022-09-10
Join: supported option to exclude columns from target data frame.

hidekoji:
# Description

supported option to exclude columns from target data frame.

# Checklist

- [ ] Pass devtools::check()
- [ ] Pass devtools::test()
- [x] Tested with Exploratory


kei51e (changes requested):
Please add test for `exclude_target_columns` argument.

kei51e (approved):

PR #1249, opened 2022-09-27
Handle NAs in conf.low/conf.high before calling approxfun

hideaki:
# Description
Fixes around Cox regression.
- Handle NAs in conf.low/conf.high before calling approxfun to avoid error.
- Avoid error from rbindlist that happens when strata column is missing for some of the PDP data frames.

# Checklist

- [ ] Pass devtools::check()
- [ ] Pass devtools::test()
- [x] Tested with Exploratory


hidekoji:
It looks good to me.

PR #1253, opened 2022-10-07
sample_df_index should return NULL (empty vector) if rate is 0 in ordered split

hideaki:
# Description
sample_df_index should return NULL (empty vector) if rate is 0 in ordered split

# Checklist

- [ ] Pass devtools::check()
- [ ] Pass devtools::test()
- [x] Tested with Exploratory

### DIFF
--- a/tests/testthat/test_randomForest_tidiers_3.R
+++ b/tests/testthat/test_randomForest_tidiers_3.R
@@ -109,6 +109,18 @@ test_that("calc_feature_imp(regression) evaluate training and test with permutat
   expect_equal(nrow(ret), 1) # 1 for train
 })
 
+test_that("calc_feature_imp() with test_split_type = ordered and test_rate = 0, which used to have an issue of creating test data set of 1 row.", {
+  set.seed(1) # For stability of result.
+  model_df <- flight %>%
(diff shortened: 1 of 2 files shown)


tests/testthat/test_randomForest_tidiers_3.R
```diff
@@ -109,6 +109,18 @@ test_that("calc_feature_imp(regression) evaluate training and test with permutat
...
+test_that("calc_feature_imp() with test_split_type = ordered and test_rate = 0, which used to have an issue of creating test data set of 1 row.", {
+  set.seed(1) # For stability of result.
+  model_df <- flight %>%
+                calc_feature_imp(`ARR DELAY`, `CAR RIER`, `ORI GIN`, `DEP DELAY`, `AIR TIME`,
+                                 test_rate = 0,
+                                 importance_measure = "firm",
+                                 test_split_type = "ordered", pd_with_bin_means = TRUE) # testing ordered split too.
+
+  ret <- rf_evaluation_training_and_test(model_df, pretty.name = TRUE)
+  expect_equal(nrow(ret), 1) # Should be 1 only for trainning set. Test set should not be created.
```
kei51e, line 121:
```suggestion
  expect_equal(nrow(ret), 1) # Should be 1 only for training set. Test set should not be created.
```

PR #1282, opened 2022-12-14
GA: Fix issue Can't subset columns that don't exist. ✖ Column `viewId` doesn't exist when there is no V3 accounts.

hidekoji:
# Description

it fixes the issue that the below error raised when there is no V3 accounts.

Error in dplyr::select(., accountId, accountName, viewId, viewName, webPropertyId, : Can't subset columns that don't exist. ✖ Column `viewId` doesn't exist.

# Checklist

- [ ] Pass devtools::check()
- [ ] Pass devtools::test()
- [x] Tested with Exploratory

### DIFF
--- a/R/google_analytics.R
+++ b/R/google_analytics.R
@@ -6,15 +6,13 @@ getGoogleProfile <- function(tokenFileId = ""){
 
   token <- getGoogleTokenForAnalytics(tokenFileId);
   googleAuthR::gar_auth(token = token, skip_fetch = TRUE)
+  # Get V3 account list.
   df <- googleAnalyticsR::ga_account_list()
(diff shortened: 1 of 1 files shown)


R/google_analytics.R
```diff
@@ -7,7 +7,15 @@ getGoogleProfile <- function(tokenFileId = ""){
   token <- getGoogleTokenForAnalytics(tokenFileId);
   googleAuthR::gar_auth(token = token, skip_fetch = TRUE)
   df <- googleAnalyticsR::ga_account_list()
```
hideaki, line 9: Add a comment saying this query is for V3.

hidekoji: Added comment.

PR #1350, opened 2023-04-21
Recode fails to replace a single dot (".") to something else and shows an error " Argument 2 must be named."

kei51e:
# Description

The `recode` function has a problem if you try to recode a dot (".") to something else. It happens only for a single dot. 

```R
> df <- tibble(text=c("a", "b", " ", "."), value=1:4)
> df %>%  dplyr::mutate(c1= dplyr::recode(text, `.` = "abc"))

Error in `dplyr::mutate()`:
  :information_source: In argument: `c1 = dplyr::recode(text, . = "abc")`.
Caused by error in `dplyr::recode()`:
  ! Argument 2 must be named.

```

It fixes this issue.

# Checklist

- [ ] Pass devtools::check()
- [ ] Pass devtools::test()
- [x] Tested with Exploratory

### DIFF
--- a/R/util.R
+++ b/R/util.R
@@ -2075,7 +2075,7 @@ setdiff <- function(x, y, force_data_type = FALSE, ...) {
 
 #'Wrapper function for dplyr::recode to workaround encoding info getting lost.
 #'@export
-recode <- function(x, ..., type_convert = FALSE) {
+recode <- function(x, type_convert = FALSE, ...) {
   # Recreate the dynamic dots for recoding "." characters.
   map <- list(...)
   ret <- dplyr::recode(x, !!!map)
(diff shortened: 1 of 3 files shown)


R/util.R
```diff
@@ -2076,7 +2076,9 @@ setdiff <- function(x, y, force_data_type = FALSE, ...) {
 #'Wrapper function for dplyr::recode to workaround encoding info getting lost.
 #'@export
 recode <- function(x, type_convert = FALSE, ...) {
-  ret <- dplyr::recode(x, ...)
+  # Recreate the dynamic dots for recoding a single dot (".") problem.
```
hideaki, line 2079:
```suggestion
  # Recreate the dynamic dots. Without it recoding a single dot (".") leads to an error when called from inside mutate().
```

kei51e: Updated comment. Thanks!

PR #1388, opened 2023-08-17
Update for the new R package versions.

kei51e:
# Description

* Fix the output column names to be consistent.
* Pass the default values instead of NULL for `seededlda::textmodel_lda` since it doesn't allow NULL anymore.

# Checklist

- [ ] Pass devtools::check()
- [x] Pass devtools::test()
- [x] Tested with Exploratory


hidekoji (approved):
It looks good to me.

PR #1389, opened 2023-08-23
Kruskal–Wallis: Fix separating comparisons column into groups.

kei51e:
# Description

Fix separating comparisons column into groups if the variable column contains the values including " - " like "1 - 2". 

# Checklist

- [ ] Pass devtools::check()
- [x] Pass devtools::test()
- [x] Tested with Exploratory


hidekoji (approved):

PR #1426, opened 2024-08-20
The broom::glance command for a survival curve model without a cohort should return the number of rows and events. 

kei51e:
# Description

The broom::glance command for a survival curve model without a cohort should return the number of rows and events. 

# Checklist

- [ ] Pass devtools::check()
- [ ] Pass devtools::test()
- [x] Tested with Exploratory


hidekoji (approved):
It looks good to me.

PR #1432, opened 2025-01-23
Enhance exp_cut_by_step function: Added validation for lower and upper range inputs to handle cases where they are outside the data range. 

kei51e:
# Description

This pull request includes changes to the `exp_cut_by_step` function in `R/system.R` to handle edge cases and improve the robustness of the function. Additionally, new test cases have been added to ensure these changes work as expected.

Improvements to `exp_cut_by_step` function:

* Added logic to handle cases where the minimum value is larger than the upper range or the maximum value is smaller than the lower range, setting these ranges to `NA` if they are invalid.
* Moved the calculation of `min.x` and `max.x` to an earlier point in the function to ensure these values are available for the new range checks.
* Removed redundant calculation of `min.x` and `max.x` after moving them to an earlier point in the function.

Addition of new test cases:

* Added test cases to verify the behavior of the function when the upper range is smaller than the minimum data range and when the lower range is larger than the maximum data range.

# Checklist

- [ ] Pass devtools::check()
- [ ] Pass devtools::test()
- [x] Tested with Exploratory

### DIFF
--- a/R/system.R
+++ b/R/system.R
@@ -4357,16 +4357,32 @@ exp_cut_by_step <- function(x, step=NA, lower.range=NA, upper.range=NA, include.
     return (cut(x, breaks=c(-Inf, Inf), include.lowest=T))
   }
 
+
+  # min/max values without NA. Those can include Inf/-Inf.
+  min.x <- min(x, na.rm=TRUE)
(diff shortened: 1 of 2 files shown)


hidekoji (approved):

PR #1517, opened 2026-06-03
feat(#30488): support auto/polychoric/mixed correlation in do_cor (16.0.14)

kei51e:
## Summary
- Add `method = "auto" | "polychoric" | "mixed"` to `do_cor()` for Analytics Correlation (#30488).
- Auto preserves factor/logical types and resolves continuous-only → Pearson, ordinal-only → Polychoric, mixed → Mixed (`polycor::hetcor`).
- Merged latest `master` (16.0.13: chi-square / Cronbach factor hardening) and bumped package version to **16.0.14**.
- `polycor` remains in Suggests (optional dependency).

## Test plan
- [ ] `devtools::test()` — especially `tests/testthat/test_stats_wrapper.R`
- [ ] Desktop #30488: Correlation with Factor Likert columns + Method=Auto succeeds (no `'arg' should be one of "pearson"...`)
- [ ] Build and upload `exploratory_16.0.14` binaries to download2 for tam pin


hidekoji (approved):

hidekoji:
Fixed the CI installation failure. `polycor` is now an optional dependency because only Polychoric/Mixed methods require it.

- Package installation no longer requires `polycor` in the CI image.
- Polychoric/Mixed now fail with a clear installation message if it is absent.
- The factor-input test skips only when the optional package is not installed.
- Verified `R CMD INSTALL --build` and the stats-wrapper tests on R 4.6.1.

hidekoji:
Fixed the key-value correlation regression reported by the package test diff. `as.data.frame()` had converted matrix column names `1..20` to `X1..X20`; the conversion now preserves names.

Verified the complete `stats_wrapper` test file (0 failures) and rebuilt/re-uploaded the macOS Intel, macOS ARM, and Windows 16.0.12 artifacts on download2.

hidekoji:
Addressed the latest CI diff: all six pre-existing Polychoric-only tests now use `skip_if_not_installed("polycor")`, matching the package's optional dependency contract. The full stats-wrapper test file passes when `polycor` is installed; the CI image will skip only those methods until it includes the optional package.

PR #1533, opened 2026-07-01
Fix test confusion matrix not updating when the classification threshold changes

kei51e:
# Description

Test-data confusion matrix and binary Summary metrics did not reflect changes to the classification threshold for tree-based classifiers (ranger / Random Forest, rpart, lightgbm, catboost). The test evaluation path used augment's precomputed `predicted_label` (fixed at the build-time threshold, effectively 0.5) instead of recomputing labels from the stored test probabilities the way the training path and xgboost already do. As a result, changing the threshold updated prediction tables but not the test confusion matrix.

## Changes

- Add internal helper `get_test_predicted_labels()` in `R/randomForest_tidiers.R` that computes threshold-aware predicted labels for test data per model class (ranger, rpart, lightgbm, catboost), consolidating the pre-existing xgboost logic. Returns `NULL` for multiclass / other model types so callers fall back to `test_pred_ret$predicted_label`.
- Wire the helper into the `conf_mat` and binary `evaluation` branches of `rf_evaluation_training_and_test()`.
- For non-xgboost models, keep `predicted_probability` from the full-length prediction output so AUC row alignment (NA-restored rows) is preserved. xgboost behavior is unchanged.

## Tests

New `tests/testthat/test_conf_mat_threshold_test_mode.R`:

- Asserts the test confusion matrix changes when the threshold changes, across Random Forest, rpart, xgboost, lightgbm, catboost, logistic regression, and binomial GLM.
- Verifies the Random Forest test confusion matrix matches the prediction table computed at the same threshold.

# Checklist

- [ ] Pass devtools::check() 
- [x] Pass devtools::test() 
- [x] Tested with Exploratory

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,8 +1,8 @@
 Package: exploratory
 Type: Package
 Title: R package for Exploratory
-Version: 15.1.22
-Date: 2026-06-20
+Version: 15.2.1
+Date: 2026-07-01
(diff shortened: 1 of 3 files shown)


hidekoji (approved):
It looks good to me.

PR #1534, opened 2026-07-02
fix: add SSL mode support to MySQL/MariaDB connections (#36505)

hidekoji:
## Description

Adds MySQL/MariaDB SSL mode handling for Exploratory issue #36505.

## Changes

- Threads `sslMode` through `queryMySQL`, `executeGenericQuery`, `getDBConnection`, `clearDBConnection`, and the column-listing error path.
- Uses RMariaDB-supported `client.flag` values instead of the unsupported `ssl.mode` argument:
  - `disable`: no SSL client flag and no CA argument
  - `allow` / `prefer`: default client behavior with the configured CA
  - `require` / `verify-ca`: `CLIENT_SSL` with the configured CA
  - `verify-full`: `CLIENT_SSL` + `CLIENT_SSL_VERIFY_SERVER_CERT` with the configured CA
- Centralizes Linux RDS CA-path normalization and MySQL/MariaDB pool-key creation so connection creation and eviction address the same connection.
- Keeps SSL mode and CA path in the pool key to prevent security-mode collisions.

RMariaDB does not expose a separate CA-only client flag, so `verify-ca` uses the supported `CLIENT_SSL` + `ssl.ca` combination.

## Tests

- Added unit coverage for all SSL modes, RMariaDB argument construction, Linux CA normalization, pool-key consistency, and mode separation.
- Verified R parsing, package installation, targeted testthat cases, and `git diff --check`.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -2,7 +2,7 @@ Package: exploratory
 Type: Package
 Title: R package for Exploratory
 Version: 15.1.24
-Date: 2026-06-20
+Date: 2026-07-01
 Authors@R: c(person("Hideaki", "Hayashi", email = "hideaki@exploratory.io", role = c("aut", "cre")), person("Hide", "Kojima", email = "hide@exploratory.io", role = c("aut")), person("Kan", "Nishida", email = "kan@exploratory.io", role = c("aut")), person("Kei", "Saito", email = "kei@exploratory.io", role = c("aut")), person("Yosuke", "Yasuda", email = "double.y.919.quick@gmail.com", role = c("aut")))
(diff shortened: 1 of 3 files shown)


DESCRIPTION
```diff
@@ -1,8 +1,8 @@
 Package: exploratory
 Type: Package
 Title: R package for Exploratory
-Version: 15.2.1
-Date: 2026-07-01
+Version: 15.1.24
+Date: 2026-06-20
```
hidekoji:
```suggestion
Date: 2026-07-01
```

kei51e (approved):

PR #1537, opened 2026-07-10
Add CHAID implementation and coverage

hidekoji:
## Summary
- Add the `chaid` implementation and export it through `NAMESPACE`
- Add test coverage for the new CHAID behavior
- Update implementation and design notes, including supporting RCA and SMOTE documentation

## Testing
- Not run (not requested)


hidekoji:
## Review: engine-level changes requested before Desktop integration (Layer A)

Reviewed against what the Exploratory Desktop Analytics framework will need to consume this engine. The engine itself is solid (self-contained, correct CHAID merging with Bonferroni, binning metadata reused at predict, graceful unknown-category traversal). Requesting the following engine-level fixes, since they change outputs that the integration layer and its tests will depend on:

### A-1. Numeric bin labels should be human-readable ranges (required)

`create_numeric_bins()` labels bins `Bin1`..`BinN`. These labels flow into `rule` strings, `edges$label`, node summary, and the category merge table — so a rule renders as `x in {Bin3 + Bin4}`, which is uninterpretable. Since interpretable rules are the primary goal of this feature, bins should be labeled from their breaks, e.g. `[18, 35]`, `(35, 52]` (with `Missing` unchanged). The labels are already persisted in `numeric_binning_map` and reused by `apply_numeric_bins()`, so only the label generation needs to change.

### A-2. Accept logical targets (required)

`validate_chaid_inputs()` rejects logical targets ("target must be character or factor"). In Exploratory, binary targets are very commonly logical columns. Suggest auto-converting logical to a factor with stable `TRUE`/`FALSE` levels instead of erroring.

### A-3. Missing warnings from the spec (should-have)

Currently only the `max_categories` skip warns. The spec also calls for warnings on: low expected cell frequencies (chi-square reliability), no significant split (root-only tree), and numeric binning applied (results depend on binning method).

### A-4. Performance validation at spec scale (before integration sign-off)

Spec targets: 10,000 rows × 20 vars comfortably; 100,000 × 50 in practical time. The pairwise category-merge loop (`merge_categories()` re-running `compute_chisq_test()` per pair per round) is the likely hotspot on high-cardinality predictors. Please add a benchmark; if slow, replacing `stats::chisq.test()` with direct `tabulate()`-based contingency counts should help.

### A-5. Minor

- `%||%` is defined locally in `R/chaid.R`; consider moving to a shared util to avoid a future collision if rlang gets imported.

hidekoji:
## Follow-up scope: Analytics framework integration layer (Layer B)

Documenting the integration work needed after this PR lands, so the engine API stays aligned with it. This PR intentionally scopes UI integration out; the following is what makes the engine consumable by the Analytics View framework (same pattern as `exp_rpart`):

### B-1. `exp_chaid()` model-df wrapper

Following the `exp_rpart` pattern in `R/randomForest_tidiers.R`:

- tidyselect NSE target + bare-column predictor selection, `target_fun` / `predictor_funs`
- framework preprocessing: `cleanup_df`, `max_nrow` sampling, `target_n` / `predictor_n` lumping into "Other" (UI path relies on this rather than the engine's `max_categories` skip)
- `test_rate` / `test_split_type` split, `seed`, group-by via `do_on_each_group`
- returns a data frame with `model` (class `c("list", ".model", ".model.chaid")`), `.test_index`, `source.data`; model carries `terms_mapping` / `orig_target_col` etc. as the evaluation framework expects

### B-2. S3 methods for `exploratory_chaid`

- `augment.exploratory_chaid` — training/test rows with `predicted_value` / `predicted_probability` (plus per-class prob columns), and `binary_classification_threshold` semantics for 2-class targets
- `tidy.exploratory_chaid` with types: `tree_nodes`, `node_summary`, `rules`, `category_merges`, `split_summary`, `conf_mat`
- `glance.exploratory_chaid` — classification metrics via the existing `evaluate_classification` helpers
- small `exploratory_chaid` branches where `rf_evaluation_training_and_test()` / `prediction()` special-case model classes

### B-3. Tree tab: reuse the interactive decision tree renderer

The interactive tree chart renders from the node data frame produced by `tidy(type = "tree_nodes")` (see `build_rpart_tree_nodes()`). The renderer is N-ary (children linked via `parent_id`), so CHAID's multiway splits render without front-end changes. `tidy.exploratory_chaid(type = "tree_nodes")` must emit the same schema: `node_id, parent_id, depth, is_leaf, edge_label, predicted, n, pct, class_json, cond_column, cond_operator, cond_value` (+ regression columns as NA). Merged category groups map to `cond_operator = "in"` with a JSON array of original categories — which is also why A-1 (range bin labels) matters here.

Two engine observations relevant to that mapping:

- `chaid_tree_data()` is close but not the consumed contract; the `tree_nodes` tidy type will be the integration point.
- `nodes$class_distribution` (named vector per node) is exactly what `class_json` needs — good.

### B-4. Tests

Model-df layer tests mirroring the existing model wrappers: wrapper happy path, group-by, test mode, logical/character/factor targets, all tidy types, evaluation pipeline end-to-end, unknown-category prediction, seed reproducibility.

hidekoji:
## Layers A + B implemented (pushed)

Addressed the two review threads above with two commits. All tests run with `pkgload::load_all()` + testthat on R 4.6.

### Layer A — engine fixes (`ba07a20`)
- **Numeric bin labels** now read as ranges (`<= 3.5`, `(3.5, 7.2]`, `> 7.2`) instead of `Bin1..BinN`, in rules, edges, node summary, and the merge table. Reused verbatim at prediction time.
- **Ordinal binned numerics**: binned numeric predictors merge only adjacent bins, so numeric split rules stay contiguous.
- **Logical targets** accepted (converted to `TRUE`-first factor levels); `target_type` recorded; numeric targets still rejected.
- **Spec warnings** emitted once per fit: numeric binning applied, no-significant-split (root-only tree), and low expected cell frequencies.
- **Performance**: category merging now builds the target-by-category contingency table once per predictor per node and evaluates every pairwise/overall chi-square as sub-matrices. Benchmarks (opt-in `tests/testthat/test_chaid_perf.R`, `EXPLORATORY_RUN_PERF=1`): 10k×20 ≈ 0.4s, 100k×50 ≈ 3s.
- Tree-growth state moved to an environment (reference semantics) to fix a lost-update bug in nested list `<<-` mutation that could collide node IDs; `%||%` moved to `util.R`.

### Layer B — Analytics integration (`205e003`)
- **`exp_chaid()`** returns the standard model data frame (`model` / `.test_index` / `source.data`, class `.model.chaid`), following `exp_rpart`: tidyselect NSE, `cleanup_df` sampling + category lumping, train/test split, seed, group-by via `do_on_each_group`.
- **`augment` / `tidy` / `glance`** S3 methods for `exploratory_chaid`, so `prediction()`, `rf_evaluation_training_and_test()`, and `tidy_rowwise()` all dispatch. `tidy` types: `evaluation`, `evaluation_by_class`, `conf_mat`, `tree_nodes`, `node_summary`, `rules`, `category_merges`, `split_summary`; `importance` errors informatively (V2).
- **`build_chaid_tree_nodes()`** emits the exact `build_rpart_tree_nodes()` schema (verified column-for-column against `exp_rpart` output), so the interactive N-ary tree renderer draws CHAID multiway splits with no front-end changes. Merged categories map to `cond_operator = "in"` with a JSON array of original categories.

### Tests
- `test_chaid.R`: 59 assertions (existing + new bin-label/adjacency/logical/warnings/counts-equivalence).
- `test_build_chaid.R`: 26 assertions — model-df shape, group-by, test mode, `tree_nodes` schema + positive-class-first ordering, `rf_evaluation_training_and_test` train+test, conf_mat, `prediction(training_and_test)`, threshold shift, report tidy types, unseen-category prediction, seed reproducibility.
- `test_chaid_perf.R`: opt-in budget guardrails.

Out of scope (V2, per the design doc): variable importance, partial dependence, weights. Desktop template + UI wiring (Layer C) is tracked separately in the tam requirements doc.

kei51e (approved):

PR #1552, opened 2026-07-21
fix(stripe): send created[gte] in fixed notation, not scientific

hidekoji:
## Description

Fixes the `Stripe - API Import` failure in the daily Mac test run:

```
Error in 'get_data(query, body)': Error Response: {
  "error": { "code": "parameter_invalid_integer",
             "message": "Invalid integer: 1.782e+09",
             "param": "created[gte]", ... } }
```

## Root Cause

`get_stripe_data()` put a raw double unixtime into the `httr` query list. `httr` composes query values with `as.character()`, and **since R 4.3 `as.character()` on a double ignores `options(scipen)`** and picks the shortest representation — so a "round" unixtime flips to scientific notation:

```r
# R 4.6.1 on the Mac test box (Mac mini M1)
as.character(1782000000)  # "1.782e+09"   <- what Stripe rejected
as.character(1781999999)  # "1781999999"
```

`date_since` is derived from `lubridate::today()`, so whether the resulting unixtime is "round" enough to flip depends on the date. That is why this looked like a flaky datasource test instead of a real bug: it only fails on the days that land on such a value. 2026-06-21 (today − 30 days) is exactly `1782000000`.

## Implementation Details

- Format the value with `sprintf("%.0f", ...)` through a small `as_integer_param()` helper (internal, not exported).
- `tests/testthat/test_get_stripe_data.R` covers it: it asserts the R-version premise (`as.character(1782000000) == "1.782e+09"`), the helper's output, and then formats **every day of the next 10 years** and asserts none come out in scientific notation.

## How to Test

```
Rscript -e '.libPaths("~/.exploratory/R/4.6_ARM"); library(testthat); library(lubridate); source("R/get_stripe_data.R"); test_file("tests/testthat/test_get_stripe_data.R")'
```
→ `[ FAIL 0 | WARN 0 | SKIP 0 | PASS 6 ]`

The end-to-end check is the daily OAuth datasource run (`test_datasource_oauth.R:216` — Stripe API Import), which needs the Stripe OAuth token on the test machine.

## Blast Radius

`get_stripe_data()` only. The value Stripe receives is unchanged for every date that already worked (`sprintf("%.0f", 1781999999) == "1781999999"`); only the scientific-notation dates change, and those were hard failures.

## Screenshots/Videos

N/A - No UI changes in this PR.


hidekoji:
Verified end-to-end against the live Stripe API on the Mac test box (read-only `GET /v1/customers`, same call as `test_datasource_oauth.R:216`), R 4.6.1 ARM:

```
unixtime = 1782000000   as.character() = 1.782e+09

BEFORE (raw double)          sent=1.782e+09    status=400  Invalid integer: 1.782e+09
AFTER  (as_integer_param)    sent=1782000000   status=200
```

And through `get_stripe_data()` itself:

```
### installed package (unpatched) -- same call as test_datasource_oauth.R:216
FAILED: Error in get_data(query, body) : Error Response: { ... "Invalid integer: 1.782e+09" ... }
### patched
rows=416 cols=50
```

The OAuth Datasource section of the daily run had no other failure — the other 20 tests all passed; Stripe was the only one.

hidekoji:
Added the preventive layer (commit 816c5ef):

**One sanitizer for every API query.** `to_api_query()` (`R/api_query.R`) formats finite numerics in fixed notation and passes everything else — strings, logicals, NA/NaN/Inf — through unchanged. All four `httr` query call sites now use it (`get_stripe_data`, `get_mailchimp_data`, and the two GitHub API calls in `system.R`); the one-off `as_integer_param()` is gone.

**Enforced, not documented.** `tests/testthat/test_api_query.R` parses the package sources, walks the AST for `httr::GET/POST/PUT/PATCH/DELETE/VERB`, and fails listing any whose `query` argument is not wrapped. Verified it catches an offender by unwrapping `get_stripe_data`:

```
Wrap these httr query arguments in to_api_query() -- an unwrapped numeric reaches the API
in scientific notation: get_stripe_data.R: query
```

So a new datasource cannot reintroduce this. `[ FAIL 0 | WARN 0 | SKIP 0 | PASS 15 ]`

**Lesson recorded** in `.claude/rules/learned.md`, including how to recognize the class: an intermittent, date-dependent API rejection echoing a value back in scientific notation is this bug — not a flaky network or an expired token.

kei51e (approved):

PR #1600, opened 2026-08-12
feat(multiple_answers): add count_na option to exp_multiple_answers_to_longer

kei51e:
## Summary
- Adds a new `exclude_na = TRUE` parameter to `exp_multiple_answers_to_longer()` that controls whether a genuinely-missing (`NA`) target-cell value is dropped, or emitted as its own answer row.
- Decouples NA handling from `exclude_empty`, which previously removed both `NA` and empty-string (`""`) choices in a single condition -- `exclude_empty` now only drops explicit empty strings; NA retention is entirely controlled by the new `exclude_na` flag.
- Bumps package version to 16.0.45.

Companion tam PR wires a new "Exclude NA" checkbox in the "Separate Multiple-Response Values into Rows" dialog through to this parameter (design doc: `docs/plans/design/37591_additional_fix_design.md` in tam).

**Note:** this parameter was originally named `count_na` (default `FALSE`, meaning "keep NA off by default"). It was renamed to `exclude_na` (default `TRUE`, inverted sense) to stay consistent with the existing `exclude_empty` parameter's naming and default-checked convention. The actual default runtime behavior (NA excluded unless the user opts in) is unchanged -- only the parameter name and boolean sense flipped. See commit `e8ff1d0d` for the rename.

## Behavior

| `exclude_na` | `exclude_empty` | NA row in output? |
|---|---|---|
| TRUE (default) | TRUE (default) | No -- same as today |
| TRUE (default) | FALSE | No (today: was kept -- see note below) |
| FALSE | TRUE | Yes |
| FALSE | FALSE | Yes |

Note: a saved step with `exclude_empty = FALSE` today keeps both NA and empty-string rows. After this change (with `exclude_na` at its new default of TRUE), that step will drop NA rows on re-run while still keeping empty-string rows. This is an intentional decoupling, confirmed acceptable with the issue owner since the feature has not shipped in a release yet.

## Test plan
- [x] 5 `testthat` cases in `tests/testthat/test_multiple_answers_to_longer.R` covering: default (NA dropped), `exclude_na = FALSE` (NA kept), `exclude_na = FALSE` + `exclude_empty = TRUE` (NA kept, empty string still dropped), `exclude_na = TRUE` (default) + `exclude_empty = FALSE` (NA dropped, empty string kept), and NA handled independently across multiple target columns.
- [x] Confirmed via TDD: all 5 tests fail against the pre-change source, then pass after the implementation.
- [x] All pre-existing tests in the same file still pass unmodified (none previously exercised `NA`).
- [x] Verified via `devtools::load_all()` + `testthat::test_file()` against the R 4.6 ARM library used by the desktop app -- all green after the rename too.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,8 +1,8 @@
 Package: exploratory
 Type: Package
 Title: R package for Exploratory
-Version: 16.0.44
-Date: 2026-08-05
+Version: 16.0.45
+Date: 2026-08-11
 Authors@R: c(person("Hideaki", "Hayashi", email = "hideaki@exploratory.io", role = c("aut", "cre")), person("Hide", "Kojima", email = "hide@exploratory.io", role = c("aut")), person("Kan", "Nishida", email = "kan@exploratory.io", role = c("aut")), person("Kei", "Saito", email = "kei@exploratory.io", role = c("aut")), person("Yosuke", "Yasuda", email = "double.y.919.quick@gmail.com", role = c("aut")))
 URL: https://github.com/exploratory-io/exploratory_func
 Description: Functions for Exploratory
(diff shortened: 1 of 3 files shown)


kei51e:
We don't need it anymore. Closing.

PR #1657, opened 2026-09-04
feat(lca): variable discrimination, class distribution, stability columns (tam#38417/#38418/#38419)

hidekoji:
## What

LCA report part 2, R side. Paired with tam PR for tam#38417 / tam#38418 / tam#38419.

### tam#38417 — Analytics Report Part 2

- **`analysis_conditions`**: `Rows Removed` carries its share of the data (`279 (27.9%)`),
  denominator `n_used + excluded_nrow`. A bare count is not interpretable — 279 rows is a
  rounding error in a million and a quarter of the sample in a thousand.
- **The 1-class baseline runs no random starts.** poLCA solves it directly with no EM
  loop, so every start is identical; the old code paid for `nrep` of them and then printed
  the number beside the multi-class rows, where it reads as a reliability diagnostic.
  `random_starts` / `best_solution_reproductions` are `NA` there, and so are
  `mean_maximum_membership_probability` and `pct_low_confidence` — both are restatements
  of "there is one class", not measurements. `minimum_class_share` stays at 100%: that one
  IS a fact about the model.
- **`class_selection` gains `reproduction_rate` and `solution_stability`.** The rate is
  reproductions / starts (the count alone cannot tell "2 of 50" from "2 of 2"); the verdict
  is `Stable` / `Possibly Unstable` / `Unstable` on the joint rule the issue pins.
  `Possibly Unstable` rather than the issue's shorter "要注意"-equivalent wording because
  both `Caution` and `Needs Attention` are already report value keys elsewhere in the app,
  mapping to different Japanese, and a value translation is looked up by the value itself.
- **`lca_profile_table` gains `model_overall_probability`** = `sum_c P(C=c) P(Y=r|C=c)`,
  and `difference` is re-based onto it. The observed share is kept as
  `overall_probability` / `observed_difference` so nothing downstream changes shape.
  Measured, not assumed: at a converged ML solution the two coincide to numerical
  precision (the likelihood equations reproduce the observed one-way margins), so this is
  a definitional fix — both sides of the subtraction now come from one model — that only
  changes printed numbers where the fit did not converge, which is exactly where the old
  mixed definition was most misleading.

### tam#38418 — Variable Discrimination

New exported `calculate_lca_variable_discrimination()` plus the `variable_discrimination`
and `variable_discrimination_pairs` tidy types: per variable, the mean over class pairs of
the Total Variation Distance between the class-conditional response distributions.

Range 0..1 regardless of category count, invariant to class labels, class order and
category order, class pairs weighted equally (deliberately NOT by class size — the question
is how different the response patterns are, not how many rows sit in each). Validates that
every `variable × class` sums to 1 within 1e-6 and that every class carries the same
category set; a missing category is an upstream misalignment, not something to fill with 0.
`NA`, never 0, for a 1-class model — 0 already means "several classes whose distributions
are identical".

### tam#38419 — Class Distribution

New `class_distribution` tidy type: counts per (variable, category, assigned class) for the
100% stacked composition chart. Rows excluded from the estimation are kept with `class = NA`
so a category answered only by excluded rows renders as a full-height (NA) band instead of
disappearing. `tidyr::complete()`d so an unobserved combination is 0, not missing.

## Tests

`tests/testthat/test_lca.R`: 137 assertions passing. 11 new `test_that` blocks cover the
issue's own Test 1–9 for the discrimination score (identical distributions → 0, complete
separation → 1, the hand-computed 0.5, the 3-class 1/3, mixed category counts, the two
error cases, label and order invariance), the NA-for-one-class rule, the out-of-range /
non-finite guards, the stability rule table, the baseline blanks, the model-based marginal,
and the two new tidy types.

Two pre-existing tests were inverted on purpose: they asserted the baseline still echoed
the configured `nrep`, which is the behaviour tam#38417 removes.

Verified end to end against the tam harness (`AnalyticsHarnessLcaValidation.test.js`, 32
live-Rserve cases) with this build installed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


kei51e (approved):

hidekoji:
Merged into PR #1656. The LCA report changes and the P1/P2 fixes now live on the #1656 branch.

PR #1681, opened 2026-09-19
feat: add build_multinom_logit for Multinomial Logistic Regression analytics (tam#37033)

hidekoji:
## Summary

R side of the new Analytics type **Multinomial Logistic Regression** (exploratory-io/tam#37033). The tam PR will follow and link here.

- New `build_multinom_logit()` (`R/build_multinom_logit.R`), built on `nnet::multinom(Hess = TRUE)`, with the same calling convention as `build_polr()`: NSE target and predictors, weight, sampling, train/test split, Repeat By.
- `reference_category`: every other category is compared with it. `NULL` or `""` means the most frequent category (ties go to the first level). It is decided on the whole data, so every Repeat By group uses the same reference. An unknown value errors and lists the valid categories.
- Target must be categorical with 3 to 20 categories. A numeric target is rejected; tam converts it before the call.
- Tidiers:
  - `tidy()`: coefficients per (category, term) with Wald SE/z/p/CI, odds ratio and base level, plus `vif`, `importance` and `partial_dependence`.
  - `glance()`: adds `Converged`.
  - `augment()`: per-class probabilities and `.fitted`.
  - `evaluate_multinom_logit()`: macro one-vs-rest metrics and Log Loss.
  - Report helpers `multinom_logit_report_basic_info()` and `multinom_logit_report_multiclass_probabilities()`.
- `MaxNWts` is sized from the design instead of nnet's default of 1000, which dummy-coded predictors exceed easily.
- If the Hessian is singular, the tidier leaves SE/z/p/CI as `NA` instead of failing.
- VIF is the generalized VIF of the predictor design matrix, so it is the same for every category comparison.
- `calc_permutation_importance_polr()` and `partial_dependence.polr_exploratory()` gain an optional `prob_fun`, so both models share them. The defaults keep `build_polr()` behaviour unchanged.
- Version bumped to 16.1.34.

The legacy step-based `build_multinom()` is untouched.

## Tests

- `tests/testthat/test_build_multinom_logit.R`: 10 tests, all passing. They cover:
  - the default and explicit reference category, including the tie rule
  - estimates and SEs checked against a direct `nnet::multinom` fit
  - errors for an unknown reference, a numeric target, 2 categories and 21 categories
  - the complex column name ``航空 会社 !"#$%&'()*+, -./:;<=>?@[]^_'{|}~ 表`` as target, with a multibyte/symbol predictor, weight, NA and test split
  - glance, VIF, importance, partial dependence, the report helpers, perfect collinearity and Repeat By
- `tests/testthat/test_build_polr.R`: 56 tests, all passing after the refactor.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


hidekoji:
tam side: exploratory-io/tam#38825

kei51e (approved):

